PR #50, opened 2023-11-22
CI: split workflow and use bot to commit formatting and cargo lock changes

Chethan-rao:


.github/workflows/CI-pr.yml
```diff
@@ -0,0 +1,135 @@
+name: CI-pr
```
SanchithHegde, line 1: You can still keep the workflow name as `CI`, but split it across two different files to run differently for different triggers.

.github/workflows/CI-pr.yml
```diff
@@ -0,0 +1,135 @@
...
+      - ready_for_review
+      - reopened
+      - synchronize
+
+jobs:
+  formatting:
+    name: Check formatting
+    runs-on: ubuntu-latest
+    steps:
+      - name: Generate a token
```
SanchithHegde, line 16: This step would have to be run only for PRs from the current repository, can't run it on PRs from forks.

.github/workflows/CI-pr.yml
```diff
@@ -0,0 +1,135 @@
...
+            git commit --message 'chore: run formatter'
+            git push
+          fi
+
+  test:
+    name: Run tests on stable toolchain
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Generate a token
```
SanchithHegde, line 64: Same here.

.github/workflows/CI-pr.yml
```diff
@@ -0,0 +1,135 @@
...
+          components: clippy
+
+      - name: Install cargo-hack
+        uses: baptiste0928/cargo-install@v2.2.0
+        with:
+          crate: cargo-hack
+
+      - uses: Swatinem/rust-cache@v2.7.0
+        with:
+          save-if: ${{ github.event_name == 'push' }}
```
SanchithHegde, line 101: This step will always be skipped then?

.github/workflows/CI-push.yml
```diff
@@ -68,7 +61,6 @@ jobs:
         run: cargo clippy --all-features --all-targets
 
       - name: Cargo hack
-        if: ${{ github.event_name == 'push' }}
```
SanchithHegde, line 71: You can then remove this condition from the cache step as well.

SanchithHegde (dismissed):

SanchithHegde (approved):

SanchithHegde (approved):

PR #57, opened 2023-12-11
feat(router): use only card number for card duplication check

ShankarSinghC:
Currently we are taking the hash of entire card details (payment method data) for card duplication check. In this case if a add card request is made for same card number but with a different card holder name, cvv or expiry it would pass the card duplication check and the card would get added even though the card number is same. 
So this pr treats the card number as a unique field and performs the duplication based on it.

Test case :-
1) Payment method create.
image
2) Payment method create for same merchant and customer, change `card_holder_name` keeping the the card number same.

image
3) Locker db still has only one card inserted in locker table.
image
4) Also there is only one entry in the hash table as well.
image

### DIFF
--- a/src/routes/data/types.rs
+++ b/src/routes/data/types.rs
@@ -13,7 +13,7 @@ use crate::error;
 
 #[derive(serde::Serialize, serde::Deserialize, Debug)]
 pub struct Card {
-    card_number: masking::StrongSecret<String>,
+    pub card_number: masking::StrongSecret<String>,
     name_on_card: Option<String>,
     card_exp_month: Option<String>,
     card_exp_year: Option<String>,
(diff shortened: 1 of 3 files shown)


src/routes/data/types.rs
```diff
@@ -13,7 +13,7 @@ use crate::error;
 
 #[derive(serde::Serialize, serde::Deserialize, Debug)]
 pub struct Card {
-    card_number: masking::StrongSecret<String>,
+    pub card_number: masking::StrongSecret<String>,
     name_on_card: Option<String>,
     card_exp_month: Option<String>,
     card_exp_year: Option<String>,
```
NishantJoshi00, line 19: nit: in the future, we shouldn't have any of the fields on `Card` as public. We should either expose an interface for accessing the data as mutation

NishantJoshi00 (approved):

PR #75, opened 2024-02-06
feat(fingerprint): add fingerprint table and corresponding db interface

ShankarSinghC:
The fingerprint table consists of three columns: `id`, `card_hash`, and `card_fingerprint`. The `insert_fingerprint` function takes the `card details` and the `hash key`. It calculates the card hash using the `hmac-sha512` algorithm. This card hash is used as a lookup in the `fingerprint` table. If a match is found, the corresponding `card_fingerprint` is returned. If no match is found, then a `nano id` for that `card_hash` is generated which will used as fingerprint for that particular card and inserted into the fingerprint table, and the corresponding fingerprint is returned in the response.


src/error/transforms.rs
```diff
@@ -80,6 +80,38 @@ impl<'a> From<&'a super::StorageError> for super::HashDBError {
...
+        match value {
+            super::StorageError::DBPoolError | super::StorageError::PoolClientFailure => {
+                Self::DBError
+            }
+            super::StorageError::FindError => Self::DBFilterError,
+            super::StorageError::DecryptionError
+            | super::StorageError::EncryptionError
+            | super::StorageError::DeleteError => Self::UnknownError,
+            super::StorageError::InsertError => Self::DBInsertError,
+            super::StorageError::NotFoundError => Self::DBFilterError,
```
NishantJoshi00, line 95:
```suggestion
            super::StorageError::FindError 
            | super::StorageError::NotFoundError => Self::DBFilterError,
            super::StorageError::DecryptionError
            | super::StorageError::EncryptionError
            | super::StorageError::DeleteError => Self::UnknownError,
            super::StorageError::InsertError => Self::DBInsertError,
```

just a minor nit

src/error/transforms.rs
```diff
@@ -80,6 +80,38 @@ impl<'a> From<&'a super::StorageError> for super::HashDBError {
...
+    fn from(value: &'a super::CryptoError) -> Self {
+        match value {
+            super::CryptoError::SerdeJsonError(_)
+            | super::CryptoError::JWError(_)
+            | super::CryptoError::InvalidData(_)
+            | super::CryptoError::EncodingError(_)
+            | super::CryptoError::NotImplemented
+            | super::CryptoError::EncryptionError
+            | super::CryptoError::DecryptionError => Self::UnknownError,
+        }
```
NishantJoshi00, line 111: why is everything mapped to unknown error?

src/storage.rs
```diff
@@ -183,3 +185,23 @@ pub trait HashInterface {
         data_hash: Vec<u8>,
     ) -> Result<types::HashTable, ContainerError<Self::Error>>;
 }
+
+///
+/// Fingerprint:
+///
+/// Interface providing functional to interface with the fingerprint table in database
```
NishantJoshi00, line 192:
```suggestion
/// Interface providing functions to interface with the fingerprint table in database
```

src/storage.rs
```diff
@@ -183,3 +185,23 @@ pub trait HashInterface {
...
+#[async_trait::async_trait]
+pub trait FingerprintInterface {
+    type Error;
+
+    async fn find_by_card_hash(
+        &self,
+        card_hash: &[u8],
```
NishantJoshi00, line 199: Can we have this as secret?

src/storage/db.rs
```diff
@@ -259,3 +261,58 @@ impl super::HashInterface for Storage {
...
+    async fn insert_fingerprint(
+        &self,
+        card: types::CardNumber,
+        hash_key: Secret<String>,
+    ) -> Result<types::Fingerprint, ContainerError<Self::Error>> {
+        let algo = HmacSha512::<1>::new(hash_key.expose().into_bytes().into());
+        let card_data = serde_json::to_vec(&card)
+            .change_error(error::FingerprintDBError::EncodingError("byte-vec"))?;
+
+        let card_hash = algo.encode(card_data)?;
```
NishantJoshi00, line 297: I feel the crypto error variant can be handled properly to give more error insights, as mentioned in one of the above comment

src/storage/types.rs
```diff
@@ -97,6 +97,24 @@ pub struct HashTable {
     pub created_at: time::PrimitiveDateTime,
 }
 
+#[derive(Debug, Clone, Identifiable, Queryable)]
+#[diesel(table_name = schema::fingerprint)]
+pub struct Fingerprint {
+    pub id: i32,
+    pub card_hash: Vec<u8>,
```
NishantJoshi00, line 104: We can have this as a Secret

src/storage/types.rs
```diff
@@ -97,6 +97,24 @@ pub struct HashTable {
...
+    pub card_fingerprint: Secret<String>,
+}
+
+#[derive(serde::Serialize, Debug)]
+pub struct CardNumber(masking::StrongSecret<String>);
+
+#[derive(Debug, Insertable)]
+#[diesel(table_name = schema::fingerprint)]
+pub(super) struct FingerprintTableNew {
+    pub card_hash: Vec<u8>,
```
NishantJoshi00, line 114: same here

src/storage/types.rs
```diff
@@ -97,6 +97,24 @@ pub struct HashTable {
...
+#[derive(Debug, Clone, Identifiable, Queryable)]
+#[diesel(table_name = schema::fingerprint)]
+pub struct Fingerprint {
+    pub id: i32,
+    pub card_hash: Vec<u8>,
+    pub card_fingerprint: Secret<String>,
+}
+
+#[derive(serde::Serialize, Debug)]
+pub struct CardNumber(masking::StrongSecret<String>);
```
NishantJoshi00, line 109: It should implement `Deserialize` and not `Serialize`

NishantJoshi00 (changes requested):

src/storage.rs
NishantJoshi00: Hey this still needs to be a secret, maybe something like `&masking::Secret<[u8]>` or `masking::Secret<&[u8]>`

NishantJoshi00 (changes requested):
LGTM

NishantJoshi00 (approved):

Chethan-rao (approved):

PR #77, opened 2024-02-14
bump: bump the rust MSRV to 1.75.0

NishantJoshi00:
## Description

This change uses the `RPITIT` feature introduced in 1.75.0, and removes the `async_trait` dependency

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2874,7 +2874,6 @@ name = "tartarus"
 version = "0.1.3"
 dependencies = [
  "argh",
- "async-trait",
  "aws-config",
  "aws-sdk-kms",
  "axum",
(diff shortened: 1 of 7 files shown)


SanchithHegde (approved):
Looks good to me!

Do you want to consider adding a CI check that also verifies compilation on the MSRV toolchain?

NishantJoshi00:
> Looks good to me!
> 
> Do you want to consider adding a CI check that also verifies compilation on the MSRV toolchain?

Agreed! That would be a great addition to CI, I think we should introduce that before pushing this change.

NishantJoshi00:
> o you want to consider adding a CI check that also verifies compilation on the MSRV toolchain?

Hey @ShankarSinghC , can we add a CI to verify if the MSRV is what we have specified?

ShankarSinghC (approved):

PR #82, opened 2024-03-04
ci: add `MSRV` toolchain check

ShankarSinghC:
add `MSRV` toolchain check


.github/workflows/CI-pr.yml
```diff
@@ -58,6 +58,59 @@ jobs:
...
+          save-if: ${{ github.event_name == 'push' }}
+
+      - name: Install cargo-hack
+        uses: baptiste0928/cargo-install@v2.2.0
+        with:
+          crate: cargo-hack
+
+      - name: Deny warnings
+        shell: bash
+        run: sed -i 's/rustflags = \[/rustflags = \[\n    "-Dwarnings",/' .cargo/config.toml
```
SanchithHegde, line 108: Since we're supporting Rust 1.75, we can move the lint configuration to the `Cargo.toml` file.

See: https://blog.rust-lang.org/2023/11/16/Rust-1.74.0.html#lint-configuration-through-cargo

As part of this change, we can remove this step (from this and the stable toolchain job), and add this snippet towards the beginning of the workflow file:

```yaml
env:
  RUSTFLAGS: "-D warnings"
```

SanchithHegde (approved):

Chethan-rao (approved):

PR #88, opened 2024-04-16
feat(ttl): add ttl to locker entries

Chethan-rao:
This PR adds a `ttl` to locker entries. Added a new column in the `locker` table which holds an optional ttl of the entry. 
When a retrieve call is made to a locker entry which has ttl expired, we throw 404. Validation is added during add card too if the ttl is invalid

### Test cases - 

1. When TTL is invalid during `/data/add`

image

2. When TTL has expired during `/data/retrieve`

image


src/utils.rs
```diff
@@ -0,0 +1,62 @@
+/// Date-time utilities.
+pub mod date_time {
+    use time::{OffsetDateTime, PrimitiveDateTime};
+
+    /// Use the well-known ISO 8601 format when serializing and deserializing an
+    /// [`Option<PrimitiveDateTime>`][PrimitiveDateTime].
+    ///
+    /// [PrimitiveDateTime]: ::time::PrimitiveDateTime
```
NishantJoshi00, line 8: add an example serialized data in the doc string for more clarity.

NishantJoshi00 (commented):
Other than that, looks good

PR #93, opened 2024-05-13
fix: address non-digit character cases in card number validation

Chethan-rao:
This PR addresses the non-digit character cases in card number validation and raises an appropriate error if fails. 
This PR also includes min and max length check for card number.

## Test cases

1. Card number containing some invalid character

image

2. `109` passes luhn but fails the card length constraint 

image

3. Invalid card number not passing luhn

image

### DIFF
--- a/src/validations.rs
+++ b/src/validations.rs
@@ -3,21 +3,30 @@ use crate::error::ApiError;
 ///
 /// Minimum limit of a card number will not deceed 8 by ISO standards
 ///
-const MIN_CARD_NUMBER_LENGTH: usize = 8;
+pub const MIN_CARD_NUMBER_LENGTH: usize = 8;
 
 ///
 /// Maximum limit of a card number will not exceed 19 by ISO standards
 ///
-const MAX_CARD_NUMBER_LENGTH: usize = 19;
+pub const MAX_CARD_NUMBER_LENGTH: usize = 19;
+
+pub fn sanitize_card_number(card_number: &str) -> Result<bool, ApiError> {
+    let card_number = card_number.split_whitespace().collect::<String>();
+
+    let is_card_number_valid = Ok(card_number.as_str())
+        .and_then(validate_card_number_chars)
+        .and_then(validate_card_number_length)
+        .map(|number| luhn(&number))?;
+
+    Ok(is_card_number_valid)
+}
 
 ///
 /// # Panics
 ///
 /// Never, as a single character will never be greater than 10, or `u8`
 ///
-pub fn validate_card_number(number: &str) -> Result<bool, ApiError> {
-    let number = number.split_whitespace().collect::<String>();
-
+pub fn validate_card_number_chars(number: &str) -> Result<Vec<u8>, ApiError> {
     let data = number.chars().try_fold(
         Vec::with_capacity(MAX_CARD_NUMBER_LENGTH),
         |mut data, character| {
@@ -35,11 +44,15 @@ pub fn validate_card_number(number: &str) -> Result<bool, ApiError> {
         },
     )?;
 
-    Ok(
-        (data.len() >= MIN_CARD_NUMBER_LENGTH && data.len() <= MAX_CARD_NUMBER_LENGTH)
-            .then(|| luhn(&data))
-            .unwrap_or(false),
-    )
+    Ok(data)
+}
+
+pub fn validate_card_number_length(number: Vec<u8>) -> Result<Vec<u8>, ApiError> {
+    if number.len() >= MIN_CARD_NUMBER_LENGTH && number.len() <= MAX_CARD_NUMBER_LENGTH {
+        Ok(number)
+    } else {
+        Err(ApiError::ValidationError("invalid card number length"))
+    }
 }
 
 #[allow(clippy::as_conversions)]
--- a/src/validations.rs
+++ b/src/validations.rs
@@ -1,7 +1,11 @@
+use crate::error::ApiError;
+
+const MIN_CARD_NUMBER_LENGTH: usize = 12;
+
 ///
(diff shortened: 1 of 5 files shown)


NishantJoshi00 (changes requested):
Don't modify the existing function create a different fallible function and then cascade the existing one with that one using something like and_then

Chethan-rao:
> Don't modify the existing function, create a different fallible function and then cascade the existing one with that one using something like and_then

We'll anyway have to modify the current function for card number length constraint. So why don't we do it on that itself. 
Also I think its better we rename the `luhn_on_string` function to something like `validate_card_number` as this function does more validation including the luhn validation?

src/validations.rs
```diff
@@ -1,29 +1,41 @@
+use crate::error::ApiError;
+
+const MIN_CARD_NUMBER_LENGTH: usize = 12;
```
SanchithHegde, line 3: Can we add doc comments for this?

src/validations.rs
```diff
@@ -1,29 +1,41 @@
...
+
 ///
 /// Maximum limit of a card number will not exceed 19 by ISO standards
 ///
-const CARD_NUMBER_LENGTH: usize = 20;
+const MAX_CARD_NUMBER_LENGTH: usize = 19;
 
 #[allow(clippy::expect_used)]
```
SanchithHegde, line 10: Let's move this `allow` annotation above that specific line/expression please.

src/validations.rs
```diff
@@ -1,29 +1,41 @@
...
         .chars()
-        .filter_map(|value| value.to_digit(10))
-        .map(|value| {
-            value
-                .try_into()
-                .expect("error while converting a single character to u8")
-        }) // safety, a single character will never be greater
-        // `u8`
-        .collect();
+        .try_fold(Vec::with_capacity(20), |mut data, character| {
```
SanchithHegde, line 21: We can maybe use the `MAX_CARD_NUMBER_LENGTH` const for specifying the capacity?

```suggestion
        .try_fold(Vec::with_capacity(MAX_CARD_NUMBER_LENGTH), |mut data, character| {
```

SanchithHegde (commented):
> [...] Also I think its better we rename the `luhn_on_string` function to something like `validate_card_number` as this function does more validation in which luhn being one validation?

+1

Should we also consider storing the sanitized card number (without whitespaces) in the database, or do we have the need to store the original value received in the request body?

Chethan-rao:
> > [...] Also I think its better we rename the `luhn_on_string` function to something like `validate_card_number` as this function does more validation in which luhn being one validation?
> 
> +1
> 
> Should we also consider storing the sanitized card number (without whitespaces) in the database, or do we have the need to store the original value received in the request body?

I think for card number its fine to store the sanitized version

SanchithHegde (dismissed):

NishantJoshi00:
Yes, but lets look at in the following way.
1. Sanitization of the data
2. Length validation
3. Luhn check

Then, won't it make sense to have the code formulated like this:

```rust
let card_number = "4242 4242 4242 4242";

fn sanitize_card_number(card_number: &str) -> Result<CardNumber> {
    Ok(card_number)
        .and_then(validate_length)
        .and_then(validate_chars)
        .and_then(luhn_on_string)
        .map(CardNumber)
}
```

Something like this, here the responsibility of each function is clear. 
1. validate_length (could be a closure): Only validates the card length
2. validate_chars: currently, being done in the `luhn_on_string`
3. luhn_on_string: performs luhn check on any string given, or better yet, discard the function and use the internal function `luhn` on `&[u8]`
4. And finally, wrapping it into the `CardNumber` type or any strong secret that we need

Chethan-rao:
> Sure, it makes sense, but I was thinking something on the lines of:
> 
> 1. Sanitization of the data
> 2. Length validation
> 3. Luhn check
> 
> Then, won't it make sense to have the code formulated like this:
> 
> ```rust
> let card_number = "4242 4242 4242 4242";
> 
> fn sanitize_card_number(card_number: &str) -> Result<CardNumber> {
>     Ok(card_number)
>         .and_then(validate_length)
>         .and_then(validate_chars)
>         .and_then(luhn_on_string)
>         .map(CardNumber)
> }
> ```
> 
> Something like this, here the responsibility of each function is clear.
> 
> 1. validate_length (could be a closure): Only validates the card length
> 2. validate_chars: currently, being done in the `luhn_on_string`
> 3. luhn_on_string: performs luhn check on any string given, or better yet, discard the function and use the internal function `luhn` on `&[u8]`
> 4. And finally, wrapping it into the `CardNumber` type or any strong secret that we need
> 
> By changing the name of the function we are just giving the function more responsibility, instead of splitting it between multiple dependent component.

Sure I have created multiple functions for each validation and are being called in `sanitize_card_number` function

SanchithHegde (approved):

src/validations.rs
```diff
@@ -1,29 +1,58 @@
...
+
+pub fn sanitize_card_number(card_number: &str) -> Result<bool, ApiError> {
+    let card_number = card_number.split_whitespace().collect::<String>();
+
+    let is_card_number_valid = Ok(card_number.as_str())
+        .and_then(validate_card_number_chars)
+        .and_then(validate_card_number_length)
+        .map(|number| luhn(&number))?;
+
+    Ok(is_card_number_valid)
```
NishantJoshi00, line 21:
```suggestion
    Ok(card_number.as_str())
        .and_then(validate_card_number_chars)
        .and_then(validate_card_number_length)
        .map(|number| luhn(&number))
```

NishantJoshi00: But, this looks very clean... good stuff!

NishantJoshi00 (approved):

PR #95, opened 2024-05-27
refactor: move crypto related managers to separate modules

Chethan-rao:
This PR moves the crypto related managers (secrets manager, encryption manager and hash managers) to separate child modules inside crypto module. 
Secrets decryption will now happen before AppState is constructed. this is a required change for multi-tenancy


src/crypto/secrets_manager/secrets_interface.rs
```diff
@@ -0,0 +1,25 @@
...
+use masking::Secret;
+
+/// Trait defining the interface for managing application secrets
+#[async_trait::async_trait]
+pub trait SecretManager: Send + Sync {
+    /// Given an input, decrypt/retrieve the secret
+    async fn get_secret(
+        &self,
+        input: Secret<String>,
+    ) -> error_stack::Result<Secret<String>, SecretsManagementError>;
```
NishantJoshi00, line 12: So, here we are considering that the return type and the input will always be a `Secret<String>` but what if in case when this isnt' true?

src/crypto/secrets_manager/managers/hcvault/implementers.rs
```diff
@@ -0,0 +1,21 @@
...
+
+#[async_trait::async_trait]
+impl SecretManager for HashiCorpVault {
+    async fn get_secret(
+        &self,
+        input: Secret<String>,
+    ) -> error_stack::Result<Secret<String>, SecretsManagementError> {
+        self.fetch::<Kv2, Secret<String>>(input.expose())
+            .await
+            .map(|val| val.expose().to_owned())
```
NishantJoshi00, line 17: what is this done?

src/crypto/secrets_manager/secrets_management.rs
```diff
@@ -0,0 +1,69 @@
...
+            #[cfg(feature = "kms-hashicorp-vault")]
+            Self::HashiCorpVault { hashi_corp_vault } => hashi_corp_vault.validate(),
+            Self::NoEncryption => Ok(()),
+        }
+    }
+
+    /// Retrieves the appropriate secret management client based on the configuration.
+    pub async fn get_secret_management_client(
+        &self,
+    ) -> error_stack::Result<Box<dyn SecretManager>, SecretsManagementError> {
```
NishantJoshi00, line 58: Minor optimization can we do this in `impl SecretManager`, this will completely eliminate the allocation as also improve the visibility on this at compile time. we can consider using enum_dispatch

NishantJoshi00 (changes requested):

src/config.rs
```diff
@@ -12,25 +12,15 @@ pub struct Config {
     pub server: Server,
     pub database: Database,
     pub secrets: Secrets,
-    #[serde(flatten)]
-    pub key_management_service: Option<EncryptionScheme>,
+    #[serde[default]]
+    pub secrets_management: SecretsManagementConfig,
```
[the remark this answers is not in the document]
Chethan-rao: I've added it in `config.example.toml`. But let it be `NoEncryption` in development.toml (by default don't have to give anything in toml for no encryption)

src/crypto/secrets_manager/secrets_interface.rs
Chethan-rao: in what case?

NishantJoshi00: In the future, might we have a usecase where we need to return let's say `Vec<u8>`?

ShankarSinghC (approved):

PR #110, opened 2024-07-10
feat: integrate a secret manager

Chethan-rao:
This PR integrates secret manager and makes use of it to store the entity dek.
Keeping in mind about the old dek's, this PR for now calls secret manager to store dek, only when new merchant hits add request. This is done to keep the number of merchant dek's to be migrated in future fixed. 
During add request, if new merchant, create a merchant dek in merchant table and also call secret manager to transfer the created dek to the service.

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -59,6 +59,17 @@ hyperswitch = { master_key = "feffe9928665731c6d6a8f9467308308feffe9928665731c6d
 # token = "hvs.abc" # The secret token to access and communicate with the vault
 
 # TLS server within axum
-# [tls]
-# certificate = "cert.pem" # path to the certificate file (`pem` format)
-# private_key = "key.pem" # path to the private key file (`pem` format)
+[tls]
+certificate = "cert.pem" # path to the certificate file (`pem` format)
+private_key = "key.pem" # path to the private key file (`pem` format)
+
+# Api client
+[api_client]
+client_idle_timeout = 90 # timeout for idle sockets being kept-alive
+pool_max_idle_per_host = 10 # maximum idle connection per host allowed in the pool.
+identity = "" # identity to be used for client certificate authentication in mtls.
+
+# Configuration for the Key Manager Service
+[key_manager]
+url = "http://localhost:5000" # URL of the encryption service
+cert = "" # Represents a server X509 certificate for mtls
--- a/src/api_client.rs
+++ b/src/api_client.rs
@@ -74,17 +74,15 @@ impl ApiClient {
 
         #[cfg(feature = "keymanager_mtls")]
         {
-            let client_identity = reqwest::Identity::from_pem(
-                global_config.api_client.identity.clone().peek().as_ref(),
-            )
-            .change_error(error::ApiClientError::IdentityParseFailed)?;
+            let client_identity =
+                reqwest::Identity::from_pem(global_config.api_client.identity.peek().as_ref())
+                    .change_error(error::ApiClientError::IdentityParseFailed)?;
 
-            let key_manager_cert = reqwest::Certificate::from_pem(
-                global_config.key_manager.cert.clone().peek().as_ref(),
-            )
-            .change_error(error::ApiClientError::CertificateParseFailed {
-                service: "key_manager",
-            })?;
+            let key_manager_cert =
+                reqwest::Certificate::from_pem(global_config.key_manager.cert.peek().as_ref())
+                    .change_error(error::ApiClientError::CertificateParseFailed {
+                        service: "key_manager",
+                    })?;
 
             client = client
                 .use_rustls_tls()
--- a/src/api_client.rs
+++ b/src/api_client.rs
@@ -1,0 +1,153 @@
+use std::str::FromStr;
+
+use crate::{
+    config::GlobalConfig,
+    error::{self, ResultContainerExt},
+};
+use error_stack::{Report, ResultExt};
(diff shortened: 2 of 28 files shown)


config.example.toml
```diff
@@ -59,6 +59,17 @@ hyperswitch = { master_key = "feffe9928665731c6d6a8f9467308308feffe9928665731c6d
...
+# Api client
+[api_client]
+client_idle_timeout = 90 # timeout for idle sockets being kept-alive
+pool_max_idle_per_host = 10 # maximum idle connection per host allowed in the pool.
+identity = "" # identity to be used for client certificate authentication in mtls.
+
+# Configuration for the Key Manager Service
+[key_manager]
+url = "http://localhost:5000" # URL of the encryption service
+cert = "" # Represents a server X509 certificate for mtls
```
NishantJoshi00, line 75: Should we take the certificate as the input? or a file path instead?

src/api_client.rs
```diff
@@ -0,0 +1,154 @@
...
+            .redirect(reqwest::redirect::Policy::none())
+            .pool_idle_timeout(std::time::Duration::from_secs(
+                global_config.api_client.client_idle_timeout,
+            ))
+            .pool_max_idle_per_host(global_config.api_client.pool_max_idle_per_host);
+
+        #[cfg(feature = "keymanager_mtls")]
+        {
+            let client_identity = reqwest::Identity::from_pem(
+                global_config.api_client.identity.clone().peek().as_ref(),
```
NishantJoshi00, line 78: Is this `.clone()` required? if we are following it with `.peek().as_ref()`

src/api_client.rs
```diff
@@ -0,0 +1,154 @@
...
+            )
+            .change_error(error::ApiClientError::IdentityParseFailed)?;
+
+            let key_manager_cert = reqwest::Certificate::from_pem(
+                global_config.key_manager.cert.clone().peek().as_ref(),
```
NishantJoshi00, line 83: same here

src/crypto/keymanager.rs
```diff
@@ -0,0 +1,134 @@
...
+                    .db
+                    .insert_entity(entity_id, &response.identifier.get_identifier())
+                    .await?)
+            }
+            false => Err(KeyManagerError::DbError.into()),
+        },
+    }
+}
+
+pub async fn transfer_key_to_key_manager(
```
NishantJoshi00, line 57: can you add comments, (as this function will be removed in the future, right?)

src/routes/key_custodian.rs
```diff
@@ -114,12 +114,15 @@ pub async fn decrypt(
...
             aes_decrypt_custodian_key(&mut tenant_config, inner_key1, inner_key2).await?;
 
-            let tenant_app_state =
-                TenantAppState::new(global_app_state.global_config.clone(), tenant_config)
-                    .await
-                    .change_context(error::ApiError::TenantError(
-                        "Failed while creating AppState for tenant",
-                    ))?;
+            let tenant_app_state = TenantAppState::new(
+                global_app_state.global_config.clone(),
```
NishantJoshi00, line 118: The global config is starting to become bulky, we should wrap it with a shared reference.

NishantJoshi00 (changes requested):
LGTM

config.example.toml
Chethan-rao: which one is convenient and secure in deployment environments? if we pass direct certificate, then we'll have to kms encrypt it. But if its file path then the file has to be kept secure. what do u suggest?

src/crypto/keymanager.rs
```diff
@@ -0,0 +1,136 @@
...
+    )
+    .await?;
+
+    Ok(response.data)
+}
+
+pub async fn decrypt_data_using_key_manager<T>(
+    state: &TenantAppState,
+    request_body: DataDecryptionRequest,
+) -> Result<T, ContainerError<KeyManagerError>>
```
NishantJoshi00, line 100: can we have it return `Secret<T>`? Its' supposedly sensitive data that is decrypted by that service, right?

src/crypto/keymanager.rs
```diff
@@ -0,0 +1,136 @@
...
+    state: &TenantAppState,
+    url: String,
+    request_body: T,
+) -> Result<R, ContainerError<E>>
+where
+    T: serde::Serialize + Send + Sync + 'static,
+    R: serde::de::DeserializeOwned,
+    ContainerError<E>: From<ContainerError<ApiClientError>> + Send + Sync,
+{
+    let headers = [("Content-type".into(), "application/json".into())]
```
NishantJoshi00, line 127: can you use the `CONTENT_TYPE` const that you have created?

src/crypto/keymanager/types.rs
```diff
@@ -0,0 +1,231 @@
...
+pub struct DataEncryptionRequest {
+    #[serde(flatten)]
+    pub identifier: Identifier,
+    pub data: DecryptedData,
+}
+
+impl DataEncryptionRequest {
+    pub fn create_request<T>(
+        key_identifier: String,
+        data: &T,
```
NishantJoshi00, line 62: can this be a `Secret` too?

NishantJoshi00 (changes requested):

dracarys18 (approved):

ShankarSinghC (approved):

NishantJoshi00 (approved):

PR #112, opened 2024-07-17
add an api for migrating old dek to keymanager and add health check for the same

Chethan-rao:
This PR adds an api for migrating the old merchant dek to keymanger.
This PR also includes keymanager health check in diagnostics endpoint

## Test cases 

1. There are 2 keys to be migrated to key manager and `entity` table is currently empty - 

image

image

2. After hitting migration endpoint - 

image

Keymanager DB - 

image

### DIFF
--- a/src/api_client.rs
+++ b/src/api_client.rs
@@ -1,0 +1,153 @@
+use std::str::FromStr;
+
+use crate::{
+    config::GlobalConfig,
+    error::{self, ResultContainerExt},
+};
+use error_stack::{Report, ResultExt};
+use masking::Maskable;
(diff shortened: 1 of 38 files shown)


NishantJoshi00:
Tests?

NishantJoshi00 (approved):

src/api_client.rs
```diff
@@ -37,6 +40,11 @@ impl HeaderExt for Headers {
     }
 }
 
+pub enum Method {
+    Get,
+    Post,
+}
```
dracarys18, line 46: No need to redefine this can use Method enum from Hyper

dracarys18 (approved):

PR #122, opened 2024-10-02
chore: include tenancy docs in setup

NishantJoshi00:
### Description

Include tenancy environment variables in `setup.md`.


Chethan-rao (changes requested):
https://github.com/juspay/hyperswitch-card-vault/blob/main/config/docker-configuration.toml

This file also needs to be updated?

Chethan-rao (approved):

ShankarSinghC (approved):

PR #123, opened 2024-10-04
feat(logging): add `console-subscriber` to support monitoring on tokio

NishantJoshi00:
### Description

Add support for `console-subscriber` making monitoring of tasks and resources more convenient.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ middleware = []
 key_custodian = []
 caching = ["dep:moka"]
 keymanager_mtls = []
+console = ["tokio/tracing", "dep:console-subscriber"]
 
 [dependencies]
 async-trait = "0.1.81"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -41,6 +42,7 @@ tracing = { version = "0.1.40" }
 tracing-appender = { version = "0.2.3" }
 tracing-attributes = "0.1.27"
 tracing-subscriber = { version = "0.3.18", default-features = true, features = ["env-filter", "json", "registry"] }
+console-subscriber = { version = "0.4.0", optional = true }
```
Chethan-rao, line 45: Does it help if this is in release too or its only for local debugging?

Chethan-rao (approved):

Cargo.toml
NishantJoshi00: It's only for local debugging, on to of that it only works if we enable these flags:

```
RUSTFLAGS="--cfg tokio_unstable"
```

ShankarSinghC (approved):

PR #139, opened 2024-12-20
refactor: record `tenant_id` and `request_id` from header in the trace span

Chethan-rao:
1. `tenant-id` and `request-id` in console logging - 

image

2. `tenant-id` and `request-id` in json logging - 

image

3. removed `map_err` in some db methods and used `change_error`, hence getting whole error stack

image

### DIFF
--- a/config/development.toml
+++ b/config/development.toml
@@ -5,7 +5,7 @@ log_format = "default"
 
 [server]
 host = "127.0.0.1"
-port = 8080
+port = 3001
 
 [limit]
 request_count = 1
@@ -54,7 +54,7 @@ Sau7H1Bhzy5G7rwt05LNpU6nFcAGVaZtzl4/+FYfYIulubYjuSEh72yuBHHyvi1/
 """
 
 [tenant_secrets]
-hyperswitch = { master_key = "feffe9928665731c6d6a8f9467308308feffe9928665731c6d6a8f9467308308", public_key = """
+public = { master_key = "feffe9928665731c6d6a8f9467308308feffe9928665731c6d6a8f9467308308", public_key = """
 -----BEGIN PUBLIC KEY-----
 MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEA5Z/K0JWds8iHhWCa+rj0
 rhOQX1nVs/ArQ1D0vh3UlSPR2vZUTrkdP7i3amv4d2XDC+3+5/YWExTkpxqnfl1T
--- a/src/error.rs
+++ b/src/error.rs
@@ -86,7 +86,7 @@ pub enum ApiError {
     #[error("Failed while decoding data")]
     DecodingError,
 
-    #[error("Failed while inserting data into \"{0}\"")]
+    #[error("Failed while inserting data into {0}")]
     DatabaseInsertFailed(&'static str),
 
     #[error("failed while deleting data from {0}")]
--- a/src/logger/formatter.rs
+++ b/src/logger/formatter.rs
@@ -126,10 +126,9 @@ where
             if !IMPLICIT_KEYS.contains(key.as_str()) {
                 true
             } else {
-                tracing::warn!(
-                    ?key,
-                    ?value,
-                    "Attempting to log a reserved entry. It won't be added to the logs"
+                eprintln!(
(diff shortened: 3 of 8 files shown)


src/error.rs
```diff
@@ -86,7 +86,7 @@ pub enum ApiError {
     #[error("Failed while decoding data")]
     DecodingError,
 
-    #[error("Failed while inserting data into \"{0}\"")]
```
NishantJoshi00, line 89: why?

src/logger/formatter.rs
```diff
@@ -179,6 +181,24 @@ where
...
 
+        // Write down entries from the span, if it exists.
+        if let Some(span) = &span {
+            let extensions = span.extensions();
+            if let Some(visitor) = extensions.get::<Storage<'_>>() {
+                for (key, value) in &visitor.values {
+                    if is_extra(key) && !explicit_entries_set.contains(key) {
+                        map_serializer.serialize_entry(key, value)?;
+                    } else {
+                        tracing::warn!(
```
NishantJoshi00, line 192: this won't cause recursion, right?

src/storage/db.rs
```diff
@@ -142,14 +148,17 @@ impl LockerInterface for Storage {
...
+            Err(err) => match err {
+                diesel::result::Error::NotFound => {
+                    Err(err).change_error(error::StorageError::NotFoundError)
+                }
+                _ => Err(err).change_error(error::StorageError::FindError),
+            },
+            Ok(locker) => Ok(locker),
+        };
+
+        output.map_err(From::from).map(|inner| inner.into())
```
NishantJoshi00, line 161:
```suggestion
        output.map_or_else(From::from, From::from)
```

NishantJoshi00 (changes requested):

src/error.rs
Chethan-rao: quotes not required right?

config/development.toml
```diff
@@ -5,7 +5,7 @@ log_format = "default"
 
 [server]
 host = "127.0.0.1"
-port = 8080
+port = 3001
```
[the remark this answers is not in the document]
Chethan-rao: since router has 8080 set, for easy and fast startup we'll keep different port

src/storage/db.rs
Chethan-rao: Here the output requires type conversion for both Ok and Err variant. map_or_else would have worked for Option

src/logger/formatter.rs
Chethan-rao: fixed it

src/logger/formatter.rs
```diff
@@ -118,11 +117,23 @@ where
...
+        mut default_fields: HashMap<String, Value>,
     ) -> Self {
         let pid = std::process::id();
         let hostname = gethostname::gethostname().to_string_lossy().into_owned();
         let service = service.to_string();
+        default_fields.retain(|key, value| {
+            if !IMPLICIT_KEYS.contains(key.as_str()) {
+                true
+            } else {
+                tracing::warn!(
```
SanchithHegde, line 129: Will we have a functional logger at this point of time, or would we have to resort to using `eprintln!()`?

Chethan-rao: ya right `eprintln!`

src/logger/storage.rs
```diff
@@ -23,11 +23,19 @@ pub struct Storage<'a> {
...
-impl Storage<'_> {
+impl<'a> Storage<'a> {
     /// Default constructor.
     pub fn new() -> Self {
         Self::default()
     }
+
+    pub fn record_value(&mut self, key: &'a str, value: serde_json::Value) {
+        if super::formatter::IMPLICIT_KEYS.contains(key) {
+            tracing::warn!("{key} is a reserved entry. Skipping it. value: {value}");
```
NishantJoshi00, line 34: is this safe to call here?

Chethan-rao: warn? yes, tested it with duplicate keys, this just logs the warn message and won't cause stack overflow

NishantJoshi00: sure

NishantJoshi00 (approved):

PR #170, opened 2026-06-21
refactor(storage): split storage into single-query primitives with a domain composition layer

ShankarSinghC:
## What

Reshapes the storage layer so each table exposes only **single-query primitives** that map 1:1 onto a future KV backend, and moves the multi-step compositions (`insert-or-get`, `find-or-create`, `upsert`) into a new `crate::domain` layer. This mirrors Hyperswitch's `storage_impl` conventions and prepares the locker/vault/merchant/fingerprint/hash data paths for a KV backend.

## Storage primitives
- `find_by_*` → return `T`, error with `NotFound` on a missing row (must-exist reads); each spells out its full key columns.
- `find_optional_by_*` → return `Option<T>` (dedup / maybe-present lookups).
- `insert_*` → surface a unique-key conflict via `StorageErrorExt::is_duplicate()`.

## Domain layer (`crate::domain`)
One composition fn per table sequencing the primitives, race-safe (insert → on duplicate → read back):
- `domain::locker::get_or_insert`
- `domain::vault::{get_or_insert, upsert}`
- `domain::hash::insert_or_get`
- `domain::merchant::find_or_create`
- `domain::fingerprint::get_or_insert`

## Error layer
- `impl_storage_error!` macro centralises raw diesel-error classification (`NotFound → NotFoundError`, `UniqueViolation → Duplicate`) — the seam where a KV `null → not-found` mapping will plug in.
- `is_duplicate()` / `is_not_found()` helpers used by the domain compositions.

## Notes
- Behaviour preserved; the `find_or_create` paths are now race-safe.
- Caching wrappers, routes, and the keymanager updated to the new method names.
- Verified: `cargo clippy --features release -- -D warnings`, default + `--no-default-features` builds, and `cargo fmt --check`.

Opening as **draft** for early review of the layering before wiring the KV backend.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

## ✅ Testing

Validated end-to-end against a running locker + Postgres, on **both** DEK backends with identical behaviour:

- **Internal key manager** — master key, no external service
- **External encryption service** — the `external_key_manager` path (`entity` table + `ExternalCryptoManager`, encrypt/decrypt over HTTP). Tested against a contract-compatible stub implementing `/key/create`, `/key/transfer`, `/data/encrypt`, `/data/decrypt`, `/health`.

Full assertion suite: **31 passed / 0 failed** on each backend.

### Cases covered

| Flow | Cases |
|---|---|
[...]

PR #172, opened 2026-06-23
feat(entity): add explicit POST /entity provisioning endpoint

ShankarSinghC:
## What

Adds a dedicated, idempotent **`POST /entity/create`** endpoint to provision the key-holder record explicitly, instead of relying on lazy creation during the add-card / add-data flow.

The backing table is chosen by configuration (same dispatch the add flow already uses):
- **internal key manager** → `merchant` table
- **external key manager** → `entity` table

So this one endpoint covers both the "create merchant" and "create entity" cases — a given deployment only ever uses one.

### Request
```json
{ "entity_id": "id_1234" }
```
### Response — 200 (idempotent; returns the existing record unchanged if present)
```json
{ "status": "Ok", "entity_id": "id_1234", "created_at": "2026-06-13 10:30:00.0" }
```

## Why now / backwards compatibility

The add flows (`/cards/add`, `/data/add`, `/api/v2/vault/add`) are **unchanged** — they still auto-create, so existing clients that haven't integrated the new endpoint keep working. **Nothing breaks.**

To make removal of the lazy fallback a data-driven decision later, the auto-create path now emits a distinct warn log **only when it actually creates** a record:

```
deprecation="add_flow_auto_create" entity_id=... "...clients should call POST /entity/create explicitly"
```

When that log rate drops to ~0 over a chosen window, a follow-up PR can: swap `find_or_create_entity` → `find_by_entity_id` in the two add handlers, delete `find_or_create_entity` from the trait + both impls, and add a changelog note (add then returns 404 if create wasn't called first).

## Changes

- `KeyProvider` trait: add `CreatedEntity` + `create_entity` (blessed provisioning path), implemented for internal and external key managers.
- Internal: `create_entity` delegates to the race-safe `domain::merchant::find_or_create`; `find_or_create_entity` reads first and warns only on actual auto-create.
- External: `create_entity` does find → create-key-in-KM → `insert_entity`; same deprecation warn added to the lazy path.
- New `src/routes/entity.rs` (request/response + handler); registered in `routes.rs` and `app.rs`.

## Notes

- Stacked on top of #170 (`refactor/storage-kv-primitives`); base is set accordingly.
- `created_at` serializes via `time`'s default format, not RFC 3339 — easy to switch if preferred.
- Verified: `cargo check` + `cargo clippy` clean on both default and `external_key_manager` feature sets.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/crypto/keymanager/internal_keymanager.rs
```diff
@@ -41,14 +42,47 @@ impl super::KeyProvider for InternalKeyManager {
...
+            .await
+        {
+            Ok(merchant) => merchant,
+            Err(err) if err.is_not_found() => {
+                logger::warn!(
+                    entity_id = %entity_id,
+                    deprecation = "add_flow_auto_create",
+                    "merchant auto-created during add flow; clients should call POST /entity explicitly"
+                );
+                merchant::find_or_create(tenant_app_state, &entity_id, &master_encryption).await?
```
hrithikesh026, line 80: We can call the dedicated find_merchant_account function. Will be easier to deprecate the existing 'find_or_create' function later.

src/routes/entity.rs
```diff
@@ -0,0 +1,80 @@
...
+        }
+    }
+}
+
+/// Response body for `POST /entity`.
+#[derive(Debug, Serialize)]
+pub struct CreateEntityResponse {
+    pub entity_id: String,
+    /// RFC 3339, UTC, millisecond precision (e.g. `2026-06-24T19:27:37.552Z`) to match the
+    /// timestamp format used by the other APIs.
```
hrithikesh026: if this timestamp format is already used by other apis, the serialize_with implementation should already exist no.

src/routes/entity.rs
```diff
@@ -0,0 +1,80 @@
...
+    #[serde(serialize_with = "serialize_rfc3339_millis")]
```
SanchithHegde: Nit: Can we consider using ISO 8601 to match the format used by Hyperswitch?

src/utils.rs
```diff
@@ -11,6 +11,24 @@ pub mod date_time {
...
+    /// with millisecond precision and a `Z` offset, e.g. `2026-06-24T19:27:37.552Z` — matching
+    /// the timestamp format used across the Hyperswitch APIs. Intended for use with
+    /// `#[serde(serialize_with = "crate::utils::date_time::serialize")]`.
+    pub fn serialize<S>(date_time: &PrimitiveDateTime, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        let format = time::macros::format_description!(
+            "[year]-[month]-[day]T[hour]:[minute]:[second].[subsecond digits:3]Z"
+        );
```
SanchithHegde: Nit: Can we consider replicating what's done in Hyperswitch (using constants from the `time` crate), rather than handwriting the format literal ourselves?

https://github.com/juspay/hyperswitch/blob/aa2beff1e15dc7958d12292485eeb8b26326e6c6/crates/common_utils/src/custom_serde.rs#L20-L47

Cargo.toml
```diff
@@ -65,7 +65,7 @@ hyperswitch_masking = { version = "0.0.1", features = ["diesel", "serde"] }
 hyperswitch_redis_interface = { git = "https://github.com/juspay/hyperswitch", tag = "2026.07.17.0", package = "redis_interface", default-features = false, features = ["fred", "metrics"], optional = true }
 ring = { version = "0.17.14", features = ["std"] }
 hex = "0.4.3"
-time = { version = "0.3.45" }
+time = { version = "0.3.45", features = ["formatting", "macros"] }
```
hrithikesh026: This can be removed

src/utils.rs
```diff
@@ -11,6 +11,24 @@ pub mod date_time {
         let utc_date_time = OffsetDateTime::now_utc();
         PrimitiveDateTime::new(utc_date_time.date(), utc_date_time.time())
     }
+
+    /// Serialize a [`PrimitiveDateTime`] (assumed to be UTC) as an ISO 8601 / RFC 3339 string
+    /// with millisecond precision and a `Z` offset, e.g. `2026-06-24T19:27:37.552Z` — matching
+    /// the timestamp format used across the Hyperswitch APIs. Intended for use with
+    /// `#[serde(serialize_with = "crate::utils::date_time::serialize")]`.
+    pub fn serialize<S>(date_time: &PrimitiveDateTime, serializer: S) -> Result<S::Ok, S::Error>
```
hrithikesh026: This is already available in main. Reuse it

hrithikesh026 (changes requested):

hrithikesh026 (dismissed):

src/routes/entity.rs
```diff
@@ -0,0 +1,62 @@
...
+    }
+}
+
+/// Response body for `POST /entity`.
+#[derive(Debug, Serialize)]
+pub struct CreateEntityResponse {
+    pub entity_id: String,
+    /// ISO 8601 UTC timestamp, using the shared locker timestamp format.
+    #[serde(serialize_with = "crate::utils::primitive_datetime_serde::iso8601::serialize")]
+    pub created_at: time::PrimitiveDateTime,
```
SanchithHegde: Nit, optional change:

You could just do:
```suggestion
    #[serde(with = "crate::utils::primitive_datetime_serde::iso8601")]
    pub created_at: time::PrimitiveDateTime,
```

In this case, it won't make a difference. If deriving `Deserialize`, you wouldn't have to explicitly specify `serde(deserialize_with = "...")` if you've specified `serde(with = "...")`.

src/crypto/keymanager/external_keymanager.rs
```diff
@@ -215,7 +216,15 @@ impl super::KeyProvider for ExternalKeyManager {
...
             Err(inner_err) => match inner_err.is_not_found() {
+                // DEPRECATED lazy provisioning: clients should call `POST /entity`
+                // explicitly. Once this warning stops appearing the fallback can be removed and
+                // the add flow switched to `find_by_entity_id`.
                 true => {
+                    logger::warn!(
+                        entity_id = %entity_id,
+                        deprecation = "add_flow_auto_create",
+                        "entity auto-created during add flow; clients should call POST /entity explicitly"
+                    );
```
SanchithHegde, line 227: Nit: You can also have a metric if you want, which would record the count of entities being created implicitly / automatically. This could also have an attribute indicating whether the internal keymanager was used or the external one.

SanchithHegde (dismissed):
Other than that, looks good to me!

SanchithHegde (approved):

PR #173, opened 2026-06-23
fix(config): fix failures in parsing metrics port from environment variables

SanchithHegde:
### Description

This PR fixes an issue in parsing metrics port from environment variables. If the application was run with a command like `LOCKER__METRICS__PORT=9090 cargo run`, we would notice the following error during application startup, which this PR fixes by calling `.try_parsing(true)`:

```text
thread 'main' (1) panicked at src/bin/locker.rs:7:61:
Failed while parsing config: invalid type: string "9090", expected u16
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Additionally, this PR includes some CI changes aimed to reduce the time taken by CI workflows.

### DIFF
--- a/.github/workflows/CI-pr.yml
+++ b/.github/workflows/CI-pr.yml
@@ -25,7 +25,7 @@ jobs:
         id: generate_token
         uses: actions/create-github-app-token@v3
         with:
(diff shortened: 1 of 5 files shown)


SanchithHegde:
CI checks pass, will be merging this PR.

PR #174, opened 2026-06-24
feat(storage): introduce reverse_lookup table

hrithikesh026:
## Summary
This PR introduces the `reverse_lookup` table to the locker, following the existing Diesel + migration patterns and aligning with the reference implementation in the main Hyperswitch repo.

## Changes
- Added a new Diesel migration to create the `reverse_lookup` table.
- Updated `src/storage/schema.rs` with the new table definition.
- Added `ReverseLookup` and `ReverseLookupNew` types in `src/storage/types.rs`.
- Introduced `ReverseLookupInterface` in `src/storage.rs`.
- Implemented the interface for `Storage` in `src/storage/db.rs`.
- Added `ReverseLookupDBError` and the corresponding error transforms.

## Verification
- `cargo check` passes
- `cargo check --all-features` passes

### DIFF
--- a/migrations/2026-06-24-125858_create_reverse_lookup_table/up.sql
+++ b/migrations/2026-06-24-125858_create_reverse_lookup_table/up.sql
@@ -4,7 +4,8 @@ CREATE TABLE IF NOT EXISTS reverse_lookup (
     lookup_id BYTEA NOT NULL PRIMARY KEY,
     secondary_key VARCHAR NOT NULL,
     partition_key VARCHAR NOT NULL,
-    source VARCHAR NOT NULL
+    source VARCHAR NOT NULL,
+    update_by VARCHAR NOT NULL
 );
 
 CREATE INDEX IF NOT EXISTS lookup_id_index ON reverse_lookup (lookup_id);
--- a/migrations/2026-06-24-125858_create_reverse_lookup_table/up.sql
+++ b/migrations/2026-06-24-125858_create_reverse_lookup_table/up.sql
@@ -2,8 +2,8 @@
 
 CREATE TABLE IF NOT EXISTS reverse_lookup (
     lookup_id BYTEA NOT NULL PRIMARY KEY,
-    sk_id VARCHAR NOT NULL,
-    pk_id VARCHAR NOT NULL,
+    secondary_key VARCHAR NOT NULL,
+    partition_key VARCHAR NOT NULL,
     source VARCHAR NOT NULL
 );
 
--- a/migrations/2026-06-24-125858_create_reverse_lookup_table/up.sql
+++ b/migrations/2026-06-24-125858_create_reverse_lookup_table/up.sql
@@ -1,7 +1,7 @@
 -- Your SQL goes here
 
 CREATE TABLE IF NOT EXISTS reverse_lookup (
-    lookup_id BYTEA NOT NULL PRIMARY KEY,
+    lookup_id VARCHAR NOT NULL PRIMARY KEY,
     secondary_key VARCHAR NOT NULL,
     partition_key VARCHAR NOT NULL,
     source VARCHAR NOT NULL,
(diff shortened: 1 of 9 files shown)


su-shivanshmathur (dismissed):

migrations/2026-06-24-125858_create_reverse_lookup_table/up.sql
```diff
@@ -0,0 +1,10 @@
+-- Your SQL goes here
+
+CREATE TABLE IF NOT EXISTS reverse_lookup (
+    lookup_id VARCHAR(255) NOT NULL PRIMARY KEY,
+    sk_id VARCHAR(50) NOT NULL,
+    pk_id VARCHAR(255) NOT NULL,
```
ShankarSinghC: Can we name it better ?

migrations/2026-06-24-125858_create_reverse_lookup_table/up.sql
```diff
@@ -0,0 +1,10 @@
+-- Your SQL goes here
+
+CREATE TABLE IF NOT EXISTS reverse_lookup (
+    lookup_id BYTEA NOT NULL PRIMARY KEY,
+    secondary_key VARCHAR NOT NULL,
+    partition_key VARCHAR NOT NULL,
+    source VARCHAR NOT NULL
+);
+
+CREATE INDEX IF NOT EXISTS lookup_id_index ON reverse_lookup (lookup_id);
```
SanchithHegde: `lookup_id` is already a primary key, do we need another index on the same column?

src/storage/types.rs
```diff
@@ -140,6 +140,36 @@ impl<'a> LockerNew<'a> {
     }
 }
 
+#[derive(Debug, Clone, Identifiable, Queryable)]
+#[diesel(table_name = schema::reverse_lookup, primary_key(lookup_id))]
+#[allow(dead_code)]
```
SanchithHegde: Nit: Let's change this to `expect(dead_code)` instead of `allow(dead_code)`. When the dead code warning is no longer emitted, the compiler would inform us and we can remove the annotation. Does not make any significant difference, only helps keep code slightly cleaner.


```suggestion
#[expect(dead_code)]
```

src/storage.rs
```diff
@@ -319,6 +319,29 @@ pub(crate) trait FingerprintInterface {
     ) -> Result<types::Fingerprint, ContainerError<Self::Error>>;
 }
 
+#[allow(dead_code)]
```
SanchithHegde: Nit:

```suggestion
#[expect(dead_code)]
```

src/storage/types.rs
```diff
@@ -140,6 +140,36 @@ impl<'a> LockerNew<'a> {
...
+pub(crate) struct ReverseLookup {
+    pub lookup_id: Vec<u8>,
```
SanchithHegde: Why is this field / column `BYTEA` / `Vec<u8>`? Are we expecting to store any lookup IDs which are not human readable?

src/storage/types.rs
```diff
@@ -140,6 +140,36 @@ impl<'a> LockerNew<'a> {
...
+    pub secondary_key: String,
+    pub partition_key: String,
+    pub source: String,
+}
```
SanchithHegde, line 152: Don't we need some field similar to what the `updated_by` field does in the Hyperswitch codebase?

src/storage/types.rs
hrithikesh026: yes. One of the reverse lookup id is of type bytea. `data_hash` column of `hash_table`.

hrithikesh026: actually, we can eliminate that reverse lookup as well. Changed this back to VARCHAR.

ShankarSinghC (approved):

PR #175, opened 2026-06-24
feat(kv): introduce kv framework

su-shivanshmathur:
# Introduce a write-through Redis KV framework

## What this does

Adds an optional Redis "KV" layer for the `fingerprint` and `hash_table` tables:

- **Reads** check Redis first, then fall back to Postgres if it's not there.
- **Writes** go to Redis and also onto a drainer stream, so a future drainer can write them
  into Postgres later.
- You turn it on **per tenant, per table** in config (`redis_kv` vs `postgres_only`).

It lives behind a new `kv` feature flag and is **off by default**, so nothing changes for
existing builds or the normal Postgres flow.

## How it's structured

- `src/storage/kv/` — the framework: run a KV op (`Get`/`SetNx`), pick the storage scheme, compute the Redis shard, serialize the SQL for the drainer, and the Redis-first / Postgres
  fallback helper.
- `src/storage/db.rs` — the `fingerprint` and `hash_table` read/insert methods now take the Redis path when that table is set to `redis_kv`, otherwise they behave exactly as before.
- Config: a global `[kv]` block (stream name, shard count, TTL) and per-table settings under `[tenant_secrets.<tenant>.kv.<table>]`, plus a per-tenant `redis_key_prefix`.
- A migration adds an `updated_by` column to both tables (tracks which path last wrote a row).

## What we reused from hyperswitch

The goal is to match hyperswitch's format exactly, so its drainer can replay our entries without changes. We **did not** import hyperswitch's big domain crates — we copied just the small, needed pieces and depend on one shared crate:

- **Dependency:** `redis_interface` (the Redis connection pool + commands).
- **Copied (adapted to card-vault):**
  - the KV executor and drainer-stream push (`kv_wrapper`, `push_to_drainer_stream`)
  - storage-scheme decision (`decide_storage_scheme`)
  - shard/partition keys (`shard_{crc32 % N}`)
  - unique-constraint checks via Redis `SADD`
  - Redis-first-then-DB helper (`try_redis_get_else_try_database_get`)
  - Redis→storage error mapping (`to_redis_failed_response`)
  - `SerializableQuery` — serializes a SQL insert for the drainer (our version needs **no** DB
    connection)
  - stream name `{shard_N}_DRAINER_STREAM` and the `redis_key_prefix` behavior

## Scope

- **In this PR:** `fingerprint` + `hash_table`, the `Get` + `SetNx` paths.
- **Later:** the drainer process itself, KV updates, and `locker`/`vault` support.

## Testing

Unit tests cover shard keys, query serialization round-trip, scheme selection, and the fallback path. `cargo clippy --features kv -- -D warnings` and `cargo check --features release` are clean. Verified end-to-end against a local Redis + Postgres.

## Notes for reviewers

- The vault uses error-stack 0.5 but `redis_interface` uses 0.4, so a small `BridgeRedis` shim converts errors between them. It can be removed once both use the same version.
- The Redis key prefix now comes from `redis_key_prefix` (empty for the `public` tenant), not the tenant id — multi-tenant setups must give each non-default tenant its own prefix.


config/config.example.toml
```diff
@@ -113,6 +113,22 @@ unresponsive_check_interval = 2    # how often to check for unresponsiveness (se
 max_feed_count = 200               # max commands fed to the server at once
 max_failure_threshold_seconds = 5  # max seconds Redis may be unreachable before it's failed
 
+# KV (write-through Redis) configuration — global drainer settings (only used with the `kv` feature)
+[kv]
+drainer_stream_suffix = "DRAINER_STREAM"  # suffix for the drainer stream name
```
hrithikesh026, line 118: Lets just call this stream name???

hrithikesh026: because prefix of this is only just the shard number.

config/development.toml
```diff
@@ -28,6 +28,14 @@ locker_private_key = ""
...
 public = { master_key = "feffe9928665731c6d6a8f9467308308feffe9928665731c6d6a8f9467308308", public_key = "", schema = "public" }
 
+[tenant_secrets.public.kv.fingerprint]
+storage_scheme = "redis_kv"
+soft_kill = false
+
+[tenant_secrets.public.kv.hash_table]
+storage_scheme = "redis_kv"
+soft_kill = false
+
```
hrithikesh026, line 38: purpose of these configs?

src/domain/fingerprint.rs
```diff
@@ -0,0 +1,49 @@
...
+    app::TenantAppState,
+    crypto::hash_manager::{hash_interface::Encode, managers::sha::HmacSha512},
+    error::{self, ContainerError, StorageErrorExt},
+    storage::{FingerprintInterface, consts, types::Fingerprint, utils},
+};
+
+/// Compute the `fingerprint_hash` for `data` (HMAC) and return the stored fingerprint,
+/// inserting a new row if none exists. The hash is the canonical dedup key, so an existing
+/// row is returned regardless of any caller-supplied `fingerprint_id`.
+pub async fn get_or_insert(
```
hrithikesh026, line 13: this change seems irrelevant for KV feature.

src/domain/hash.rs
```diff
@@ -0,0 +1,26 @@
...
+    app::TenantAppState,
+    error::{self, ContainerError, StorageErrorExt},
+    storage::{HashInterface, types::HashTable},
+};
+
+/// Insert a hash row for `data_hash`, or return the existing one on a duplicate conflict.
+///
+/// `add_card` already checks `find_by_data_hash` first; this provides the race-safe insert
+/// for the not-found branch (`db.insert` → on duplicate → `db.find`).
+pub async fn insert_or_get(
```
hrithikesh026, line 13: this also seems irrelevant.

src/domain/locker.rs
```diff
@@ -0,0 +1,32 @@
...
+        LockerInterface,
+        types::{Locker, LockerNew},
+    },
+};
+
+/// Insert the locker row, or return the existing one on a duplicate-key conflict.
+///
+/// This is the composition that used to live inside the storage layer as
+/// `insert_or_get_from_locker`: `db.insert` → on duplicate → `db.get`.
+pub async fn get_or_insert(
```
hrithikesh026, line 14: this as well.

PR #176, opened 2026-06-30
refactor: change hash_table primary key to data_hash

hrithikesh026:
## Summary
- add Diesel migrations to create replacement unique indexes and swap hash_table primary key from hash_id to data_hash
- keep hash_id protected with a unique constraint after the swap
- update Diesel schema to use data_hash as the hash_table primary key

## Verification
- diesel migration run
- diesel migration redo
image

### DIFF
--- a/migrations/2026-06-30-000000_hash-table-hash-id-unique-index/metadata.toml
+++ b/migrations/2026-06-30-000000_hash-table-hash-id-unique-index/metadata.toml
@@ -1,2 +1,2 @@
-# Needed because postresql does not allow 'DROP/CREATE INDEX CONCURRENTLY' inside a transaction block
+# Needed because postgresql does not allow 'DROP/CREATE INDEX CONCURRENTLY' inside a transaction block
 run_in_transaction = false
--- a/migrations/2026-06-30-000000_hash-table-hash-id-unique-index/metadata.toml
+++ b/migrations/2026-06-30-000000_hash-table-hash-id-unique-index/metadata.toml
@@ -1,0 +1,2 @@
+# Needed because postresql does not allow 'DROP/CREATE INDEX CONCURRENTLY' inside a transaction block
+run_in_transaction = false
--- a/migrations/2026-06-30-000001_hash-table-data-hash-primary-index/metadata.toml
+++ b/migrations/2026-06-30-000001_hash-table-data-hash-primary-index/metadata.toml
@@ -1,2 +1,2 @@
-# Needed because postresql does not allow 'DROP/CREATE INDEX CONCURRENTLY' inside a transaction block
+# Needed because postgresql does not allow 'DROP/CREATE INDEX CONCURRENTLY' inside a transaction block
 run_in_transaction = false
--- a/migrations/2026-06-30-000001_hash-table-data-hash-primary-index/metadata.toml
+++ b/migrations/2026-06-30-000001_hash-table-data-hash-primary-index/metadata.toml
@@ -1,0 +1,2 @@
+# Needed because postresql does not allow 'DROP/CREATE INDEX CONCURRENTLY' inside a transaction block
+run_in_transaction = false
--- a/migrations/2026-06-30-000002_hash-table-primary-key-swap/down.sql
+++ b/migrations/2026-06-30-000002_hash-table-primary-key-swap/down.sql
@@ -9,6 +9,7 @@ ALTER TABLE hash_table
     ADD CONSTRAINT hash_table_data_hash_key
         UNIQUE (data_hash);
 
+-- In production environments with live traffic, run these index creations with CONCURRENTLY.
 CREATE UNIQUE INDEX IF NOT EXISTS hash_table_hash_id_key_idx
 ON hash_table (hash_id);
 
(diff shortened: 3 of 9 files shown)


migrations/2026-06-30-000002_hash-table-primary-key-swap/down.sql
```diff
@@ -0,0 +1,16 @@
...
+ALTER TABLE hash_table
+    DROP CONSTRAINT hash_table_pkey,
+    DROP CONSTRAINT hash_table_hash_id_key,
+    ADD CONSTRAINT hash_table_pkey
+        PRIMARY KEY (hash_id),
+    ADD CONSTRAINT hash_table_data_hash_key
+        UNIQUE (data_hash);
+
+CREATE UNIQUE INDEX IF NOT EXISTS hash_table_hash_id_key_idx
+ON hash_table (hash_id);
```
ShankarSinghC, line 14: won't this build the index from scratch ?
should we consider doing it `CONCURRENTLY` ?

ShankarSinghC, line 14: Also do we need this as the above query is already making hash_id as `PRIMARY KEY` ?

migrations/2026-06-30-000000_hash-table-hash-id-unique-index/metadata.toml
```diff
@@ -0,0 +1,2 @@
+# Needed because postresql does not allow 'DROP/CREATE INDEX CONCURRENTLY' inside a transaction block
```
ShankarSinghC: nit: 
```suggestion
# Needed because postgresql does not allow 'DROP/CREATE INDEX CONCURRENTLY' inside a transaction block
```

migrations/2026-06-30-000001_hash-table-data-hash-primary-index/metadata.toml
```diff
@@ -0,0 +1,2 @@
+# Needed because postresql does not allow 'DROP/CREATE INDEX CONCURRENTLY' inside a transaction block
```
ShankarSinghC: nit: 
```suggestion
# Needed because postgresql does not allow 'DROP/CREATE INDEX CONCURRENTLY' inside a transaction block
```

migrations/2026-06-30-000002_hash-table-primary-key-swap/down.sql
hrithikesh026: this is `down.sql` Can be done if needed

hrithikesh026: this is just for reversing the state of DB to the previous. I just strictly reverts all the changes done by up.sql

hrithikesh026: Actually I tried it. Doesn't work. when multiple sql queries are involved, diesel always run them in a  transaction. And concurrently cannot be used in transaction block.

migrations/2026-06-30-000002_hash-table-primary-key-swap/up.sql
```diff
@@ -0,0 +1,10 @@
+SET LOCAL lock_timeout = '2s';
+SET LOCAL statement_timeout = '5s';
```
SanchithHegde, line 2: Are we necessarily required to set these, both here and in the `down.sql` file?

migrations/2026-06-30-000002_hash-table-primary-key-swap/down.sql
SanchithHegde: > Actually I tried it. Doesn't work. when multiple sql queries are involved, diesel always run them in a transaction. And concurrently cannot be used in transaction block.

You've already added a `metadata.toml` file in other migration directories in your PR, instructing `diesel` CLI to not run those in a transaction. You can choose to do the same here, if we are absolutely required to have this index `hash_table_hash_id_key_idx` when reverting.

src/storage/schema.rs
```diff
@@ -21,7 +21,7 @@ diesel::table! {
 }
 
 diesel::table! {
-    hash_table (hash_id) {
+    hash_table (data_hash) {
         id -> Int4,
```
SanchithHegde, line 25: We should consider stopping the application from querying the `id` column (in other words, remove it from `schema.rs` file and respective Rust structs) in all tables, using `diesel`'s `patch_file` field, similar to how we did it in the Hyperswitch codebase. We can take this up in a separate PR.

migrations/2026-06-30-000002_hash-table-primary-key-swap/up.sql
hrithikesh026: while running on a production like environment with live traffic, this provides a safe way to run the below alter queries. For local development, it is not needed.

migrations/2026-06-30-000002_hash-table-primary-key-swap/down.sql
hrithikesh026: Yes. In order to bring the DB state to exactly like before this migration, we need this query. But 'concurrently' keyword is not possible. Also since these up.sql migrations are forwards compatible, I don't see why we would ever want to run down.sql. 

If need I can mention in a comment to use `CONCURRENTLY` while running in a production environment. What you think?

SanchithHegde: > If need I can mention in a comment to use `CONCURRENTLY` while running in a production environment. What you think?

Yeah we can add this comment in the `down.sql` file.

SanchithHegde (approved):

ShankarSinghC (approved):

PR #177, opened 2026-06-30
feat(kv): KV framework introduction + fingerprint table movement for kv

su-shivanshmathur:
## Summary

Adds a `kv` cargo feature with a generic KV (write-through Redis) framework. Migrates `fingerprint` as the first consumer. Other tables stay on plain Postgres.

## Changes

- **`kv` feature** — gates all KV code; adds `crc32fast`, `strum`, `error-stack-04`; bumps `hyperswitch_redis_interface` to `2026.06.25.0`
- **`StorageScheme` enum** — `PostgresOnly` / `RedisKv`; always compiled; diesel `ToSql`/`FromSql` for new `updated_by` column on all 4 tables
- **KV framework** (`src/storage/kv/`):
  - `resource.rs` — `StorageResource` trait, `insert_plain_resource` (SetNx + drainer), `find_optional_plain_resource` (Get → PG fallback)
  - `wrapper.rs` — `KvOperation`/`KvResult`, `kv_wrapper` executor, `push_to_drainer_stream`, error-stack 0.4→0.5 bridge
  - `serializable_query.rs` — serializes diesel INSERT queries to JSON for drainer replay
  - `constraints.rs` — Redis SADD dedup guard
  - `partition_key.rs` — `PartitionKey::Fingerprint`, CRC32 hash-slot routing
  - `scheme.rs` — `decide_storage_scheme` with soft-kill (Insert→PG, Find→Redis)
  - `impls.rs` — `FingerprintTableNew` trait impls
  - `metrics.rs` — 4 KV counters
- **Runtime config** — `RuntimeConfigItem` trait, generic `get::<T>()`, background prefetch task
- **KV enablement** — `locker.enable_kv` runtime-config key; TOML `[kv]` fallback; `soft_kill` gated on `enable_kv`
- **Per-request tracing** — `x-request-id` flows into drainer stream entries via `with_request_id()`
- **Fingerprint migration** — find → Redis `Get` → PG fallback; insert → Redis `SetNx` + SADD dedup + drainer stream
- **Errors** — `KVError`, `DuplicateValue`, `ValueNotFound`, `SerializationFailed`; `RedisErrorExt` mapper; transforms for all table error types
- **Config** — `[kv]` section (`drainer_stream_suffix`, `drainer_num_partitions`, `ttl_for_kv`, `enable_kv`, `soft_kill`); `redis_key_prefix` in `[tenant_secrets]`

## Write-through flow

```
Insert:  SADD unique_constraint → SETNX key (TTL) → push INSERT to drainer stream
Find:    GET key → hit: return │ miss: PG SELECT → return
Soft-kill: Insert → PG only │ Find → Redis first, PG fallback
```

## Migration notes

- Add `updated_by VARCHAR NOT NULL DEFAULT 'postgres_only'` to `fingerprint`, `hash_table`, `locker`, `vault`
- Set `locker.enable_kv` = `{"enable_kv": true, "soft_kill": false}` (or `[kv] enable_kv = true` in TOML)
- Requires a drainer consumer process reading `{shard_N}_DRAINER_STREAM`

## Test evidence (fingerprint)

`USE_JWE=true`, `enable_kv=true`, `soft_kill=false`. Redis flushed + PG truncated before run.

| Scenario | Redis | Drainer | PG | ✓ |
|---|---|---|---|---|
| Insert new fingerprint | `SETNX key → updated_by=redis_kv` | `{shard_43}` 1 INSERT entry | 1 row, `updated_by=redis_kv` | ✓ |
| Same fingerprint again | `GET` hit, no new `SETNX` | 0 new entries | 1 row (unchanged) | ✓ |
| Soft-kill mode | Insert → PG, Find → Redis-first | — | — | ✓ |

**Redis KV key:**
```
fingerprint_e581caef4cd84851a320e3aa20078d2b12652a2eab38fd3afe4709aae2d63c06e0a0a61ab5b17c8ec456a771b8274f09342269fc251db4328f55100a7c4cac37
→ {"fingerprint_hash":[229,129,202,239,76,216,72,81,163,32,227,170,32,7,141,43,18,101,42,46,171,56,253,58,254,71,9,170,226,214,60,6,224,160,166,26,181,177,124,142,196,86,167,113,184,39,79,9,52,34,105,252,37,29,180,50,143,85,16,10,124,76,172,55],"fingerprint_id":"LIxMY2AO7sUWvohWfGBJ","updated_by":"redis_kv"}
```

**Drainer stream entry** (`{shard_43}_DRAINER_STREAM`):
```
1782811316050-0
query
{"sql":"INSERT INTO \"fingerprint\" (\"fingerprint_hash\", \"fingerprint_id\", \"updated_by\") VALUES ($1, $2, $3)","binds":["5YHK70zYSFGjIOOqIAeNKxJlKi6rOP06/kcJquLWPAbgoKYatbF8jsRWp3G4J08JNCJp/CUdtDKPVRAKfEysNw==","TEl4TVkyQU83c1VXdm9oV2ZHQko=","cmVkaXNfa3Y="],"metadata":[[17,1001],[25,1009],[25,1009]],"safe_to_cache_prepared":false,"entity_type":"fingerprint","operation":"insert"}
global_id
fingerprint_e581caef4cd84851a320e3aa20078d2b12652a2eab38fd3afe4709aae2d63c06e0a0a61ab5b17c8ec456a771b8274f09342269fc251db4328f55100a7c4cac37
request_id
019f17d5-ce43-7400-8dc0-3abcfe21226e
pushed_at
1782811315
```


src/config.rs
```diff
@@ -127,6 +129,12 @@ pub struct TenantSecrets {
 
     /// schema name for the tenant (defaults to tenant_id)
     pub schema: String,
+
+    /// Redis key prefix for this tenant.  Empty (default) ⇒ keys are not
+    /// namespaced, matching hyperswitch's public-tenant convention.
+    #[cfg(feature = "redis")]
+    #[serde(default)]
+    pub redis_key_prefix: String,
```
hrithikesh026, line 147: this should go in `pub struct TenantConfig`

src/storage.rs
```diff
@@ -46,8 +87,17 @@ pub struct Storage {
     primary_pg_pool: Arc<Pool<AsyncPgConnection>>,
     replica_pg_pool: Option<Arc<Pool<AsyncPgConnection>>>,
     runtime_config_manager: Arc<crate::runtime_config::RuntimeConfigManager>,
+    #[cfg(feature = "kv")]
+    redis: Option<redis_store::RedisStore>,
+    #[cfg(feature = "kv")]
+    kv_config: crate::config::KvConfig,
+    #[cfg(feature = "kv")]
+    request_id: Option<String>,
```
hrithikesh026: do we need kv_config and request_id in Storage struct?

src/storage/caching.rs
```diff
@@ -172,6 +172,19 @@ where
...
 }
 
+/// Forward `with_request_id` through the `Caching` wrapper so that
+/// `TenantStateResolver` can set the request ID regardless of whether
+/// caching is active.
+#[cfg(feature = "kv")]
+impl Caching<super::Storage> {
+    pub fn with_request_id(self, request_id: impl Into<String>) -> Self {
+        Self {
+            inner: self.inner.with_request_id(request_id),
```
hrithikesh026: recursive call????

src/storage/db.rs
```diff
@@ -261,33 +266,73 @@ impl super::FingerprintInterface for Storage {
...
         fingerprint_hash: Secret<Vec<u8>>,
         fingerprint_id: Secret<String>,
     ) -> Result<types::Fingerprint, ContainerError<Self::Error>> {
-        let mut conn = self.get_conn().await?;
+        #[cfg(feature = "kv")]
+        {
+            let model = types::FingerprintTableNew {
+                fingerprint_hash: fingerprint_hash.clone(),
+                fingerprint_id: fingerprint_id.clone(),
+                updated_by: StorageScheme::PostgresOnly,
```
hrithikesh026, line 313:
```suggestion
                updated_by: StorageScheme::RedisKv,
```

src/storage/scheme.rs
```diff
@@ -0,0 +1,101 @@
...
+}
+
+impl FromSql<Text, Pg> for StorageScheme {
+    fn from_sql(bytes: <Pg as Backend>::RawValue<'_>) -> deserialize::Result<Self> {
+        let s = <String as FromSql<Text, Pg>>::from_sql(bytes)?;
+        s.parse::<Self>()
+            .map_err(Box::<dyn std::error::Error + Send + Sync>::from)
+    }
+}
+
```
hrithikesh026: should not have manual string conversions. Check how DB enums are handled for VARCHAR columns in hyperswitch

src/storage/types.rs
```diff
@@ -136,6 +140,7 @@ impl<'a> LockerNew<'a> {
             enc_data,
             hash_id,
             ttl: *request.ttl,
+            updated_by: StorageScheme::PostgresOnly,
```
hrithikesh026, line 144: db migrations for this is missing

src/config.rs
```diff
@@ -203,7 +211,9 @@ impl GlobalConfig {
             .add_source(
                 config::Environment::with_prefix("LOCKER")
                     .separator("__")
-                    .try_parsing(true),
+                    .try_parsing(true)
+                    .list_separator(",")
+                    .with_list_parse_key(RUNTIME_CONFIG_KEYS_PATH),
```
hrithikesh026: why is this needed?

src/config.rs
```diff
@@ -365,6 +375,37 @@ impl GlobalConfig {
...
+    /// TTL (seconds) for keys written to Redis KV.
+    pub ttl_for_kv: u32,
+    /// File-based fallback for enabling KV when `runtime_config.mode` is
+    /// `"disabled"`.  When the runtime config endpoint is enabled, the
+    /// endpoint's `locker.enable_kv` value takes precedence.
+    #[serde(default)]
+    pub enable_kv: bool,
+    /// File-based fallback for soft-kill mode (gradual rollout).
+    #[serde(default)]
+    pub soft_kill: bool,
```
hrithikesh026: do we need this when we already have a runtime config fetch?

src/config.rs
```diff
@@ -415,6 +463,12 @@ pub enum RuntimeConfig {
         ttl_seconds: u64,
         #[serde(default = "default_runtime_config_cache_max_capacity")]
         cache_max_capacity: u64,
+        /// Keys to prefetch/warm at startup and on each periodic refresh tick.
```
hrithikesh026: config refresh interval also should be a part of this

src/config.rs
```diff
@@ -415,6 +463,12 @@ pub enum RuntimeConfig {
...
+        ///
+        /// Env-overridable via `LOCKER__RUNTIME_CONFIG__KEYS=key1,key2`.
+        /// Absent keys are simply not prewarmed — `get::<T>()` still lazy-fetches.
+        #[serde(default)]
+        keys: Vec<String>,
```
hrithikesh026: Ideally we should have a single config fetch endpoint. All configs must be returned in that single api call.

src/custom_extractors.rs
```diff
@@ -28,7 +28,29 @@ impl FromRequestParts<Arc<GlobalAppState>> for TenantStateResolver {
...
+            if let Some(req_id) = parts
+                .headers
+                .get(consts::X_REQUEST_ID)
+                .and_then(|h| h.to_str().ok())
+            {
+                app_state.db = app_state.db.clone().with_request_id(req_id);
+            }
+        }
+
+        Ok(Self(Arc::new(app_state)))
```
hrithikesh026: both sections compile when kv is enabled. Not be not kv

src/error.rs
```diff
@@ -68,6 +72,17 @@ pub enum StorageError {
...
+    DuplicateValue {
+        entity: &'static str,
+        key: Option<String>,
+    },
+    #[error("KV error")]
+    KVError,
+    #[error("ValueNotFound: {0}")]
+    ValueNotFound(String),
+    #[error("Serialization failure")]
+    SerializationFailed,
```
hrithikesh026: these error types should not be included here.

src/storage.rs
```diff
@@ -36,8 +41,44 @@ pub struct ReplicaRouting {
...
+/// Per-field `#[serde(default)]` is required so partial payloads like
+/// `{"enable_kv": true}` deserialize without erroring on missing `soft_kill`.
+#[cfg(feature = "kv")]
+#[derive(Clone, Debug, Default, serde::Deserialize)]
+pub struct KvRuntimeConfig {
+    #[serde(default)]
+    enable_kv: bool,
+    #[serde(default)]
+    soft_kill: bool,
+}
```
hrithikesh026: Would be better to use an enum. 

```
struct KvRuntimeConfig{
    kv_state: KvState
}
enum KvState {
    Disabled,
    Enabled,
    SoftKill
}
```

Because a state with enable_kv=false and soft_kill=true will be confusing.

src/storage/kv/resource.rs
```diff
@@ -0,0 +1,140 @@
...
+
+    if matches!(scheme, StorageScheme::RedisKv) {
+        let result = kv_wrapper::<M, M>(store, KvOperation::<M>::Get, partition_key.clone()).await;
+
+        if let Ok(KvResult::Get(v)) = result {
+            return Ok(Some(M::into_domain(v)));
+        }
+    }
+
+    M::storage_find_optional(store, &partition_key).await
```
hrithikesh026: lets do if else instead of early return.

src/storage/kv/resource.rs
```diff
@@ -0,0 +1,140 @@
...
+
+        let drainer_query = model.insert_drainer_query().map_err(|e| {
+            kv_write_error::<M::Error>(KvWriteError::Backend(
+                e.change_context(StorageError::KVError),
+            ))
+        })?;
+
+        let reply = kv_wrapper::<(), M>(
+            store,
+            KvOperation::SetNx(&model, drainer_query),
```
hrithikesh026: we should be inserting the converted object. Not the New object. And the object that is inserted must be returned from the function.

src/storage/kv/wrapper.rs
```diff
@@ -0,0 +1,217 @@
...
+
+                if let Err(err) = value.check_for_constraints(&redis_conn).await.bridge() {
+                    if matches!(err.current_context(), RedisError::SetAddMembersFailed) {
+                        return Ok(KvResult::SetNx(SetnxReply::KeyNotSet));
+                    }
+                    return Err(err);
+                }
+
+                let result = redis_conn
+                    .serialize_and_set_key_if_not_exist(&key.into(), value, Some(i64::from(ttl)))
```
hrithikesh026:
```suggestion
                    .serialize_and_set_hash_field_if_not_exist(&key.into(), value, Some(i64::from(ttl)))
```

src/storage/kv/wrapper.rs
```diff
@@ -0,0 +1,217 @@
...
+                    push_to_drainer_stream::<S>(store, query, partition_key).await?;
+                    Ok(KvResult::SetNx(result))
+                } else {
+                    Ok(KvResult::SetNx(SetnxReply::KeyNotSet))
+                }
+            }
+
+            KvOperation::Get => {
+                let result = redis_conn
+                    .get_and_deserialize_key(&key.into(), type_name)
```
hrithikesh026:
```suggestion
                    .get_hash_field_and_deserialize(&key.into(), type_name)
```

src/storage/kv/wrapper.rs
```diff
@@ -0,0 +1,217 @@
...
+
+                if matches!(result, SetnxReply::KeySet) {
+                    push_to_drainer_stream::<S>(store, query, partition_key).await?;
+                    Ok(KvResult::SetNx(result))
+                } else {
+                    Ok(KvResult::SetNx(SetnxReply::KeyNotSet))
+                }
+            }
+
+            KvOperation::Get => {
```
hrithikesh026:
```suggestion
            KvOperation::HGet => {
```

src/storage/kv/wrapper.rs
```diff
@@ -0,0 +1,217 @@
...
+    let key = format!("{partition_key}");
+
+    let type_name = std::any::type_name::<T>();
+    let operation = op.to_string();
+
+    let ttl = store.ttl_for_kv();
+
+    let result = async {
+        match op {
+            KvOperation::SetNx(value, query) => {
```
hrithikesh026:
```suggestion
            KvOperation::HSetNx(value, query) => {
```

src/storage/kv/resource.rs
```diff
@@ -0,0 +1,140 @@
...
+    decide_storage_scheme(settings, op).await
+}
+
+#[instrument(skip(store, partition_key), fields(resource = M::ENTITY_TYPE))]
+pub(crate) async fn find_optional_plain_resource<M>(
+    store: &Storage,
+    partition_key: PartitionKey<'_>,
+) -> Result<Option<M::Domain>, ContainerError<M::Error>>
+where
+    M: StorageResource + PlainKeyed + KvFindOptional,
```
hrithikesh026: probably don't need this 

```suggestion
    M: PlainKeyed + KvFindOptional,
```

src/runtime_config.rs
```diff
@@ -120,12 +137,89 @@ impl RuntimeConfigManager {
...
+                        "Failed to prefetch runtime config key, continuing"
+                    );
+                }
+            }
+        }
+    }
+
+    /// Spawn a background task that prefetches keys at startup and refreshes
+    /// them at `ttl * 0.8`.  Returns `None` when runtime config is `Disabled`.
+    pub fn spawn_prefetch_task(self: &Arc<Self>) -> Option<tokio::task::JoinHandle<()>> {
```
hrithikesh026, line 122: Do we need this If we make sure that the Mocha cache expiry is set to a short time?

src/storage/types.rs
su-shivanshmathur: Added

src/storage/scheme.rs
su-shivanshmathur: Have tried to remove that, there could have been a way to import `router_derive` to use the proc-macro which handles the conversion on hyperswitch side 

But that would just increase the dependency on hyperswitch, so tried to add `FromSqlRow` derivation to `StorageScheme`

src/error/kv.rs
```diff
@@ -0,0 +1,38 @@
...
+            RedisError::NotFound => self.change_context(KvError::ValueNotFound(format!(
+                "Data does not exist for key {key}",
+            ))),
+            RedisError::SetNxFailed | RedisError::SetAddMembersFailed => {
+                self.change_context(KvError::DuplicateValue {
+                    entity: "redis",
+                    key: Some(key.to_string()),
+                })
+            }
+            _ => self.change_context(KvError::Backend),
```
hrithikesh026: Lets not use `_`

su-shivanshmathur: Sure, updating it

src/error/kv.rs
```diff
@@ -0,0 +1,72 @@
...
+
+impl RedisErrorExt for error_stack::Report<RedisError> {
+    fn to_redis_failed_response(self, key: &str) -> error_stack::Report<KvError> {
+        match self.current_context() {
+            RedisError::NotFound => self.change_context(KvError::ValueNotFound(format!(
+                "Data does not exist for key {key}",
+            ))),
+            RedisError::SetNxFailed | RedisError::SetAddMembersFailed => {
+                self.change_context(KvError::DuplicateValue {
+                    entity: "redis",
```
su-shivanshmathur: Should this be a constant ?

src/storage/kv/resource.rs
```diff
@@ -0,0 +1,149 @@
...
+        .await;
+
+        if let Ok(KvResult::HGet(v)) = result {
+            Ok(Some(M::into_domain(v)))
+        } else {
+            // Redis miss or error — fall through to Postgres.
+            M::storage_find_optional(store, &partition_key).await
+        }
+    } else {
+        M::storage_find_optional(store, &partition_key).await
```
su-shivanshmathur: Can this be in a match block ?

src/storage/kv/impls.rs
```diff
@@ -0,0 +1,101 @@
...
+        Storage,
+        types::{Fingerprint, FingerprintTableNew},
+    },
+};
+
+impl EntityType for FingerprintTableNew {
+    const ENTITY_TYPE: &'static str = "fingerprint";
+}
+
+impl KvStorePartition for Fingerprint {}
```
su-shivanshmathur: Is this needed ?

src/error/kv.rs
hrithikesh026: entity here should be table name.

src/storage/kv/resource.rs
```diff
@@ -0,0 +1,144 @@
...
+/// DB find op for KV-routed tables. `StorageResource` is a super-trait bound.
+pub(crate) trait KvFindOptional: StorageResource {
+    async fn storage_find_optional(
+        store: &Storage,
+        pk: &PartitionKey<'_>,
+    ) -> Result<Option<Self::Domain>, ContainerError<Self::Error>>;
+}
+
+/// Marker for plain-keyed tables (Redis HGet/HSetNx).
+pub(crate) trait PlainKeyed {}
```
hrithikesh026: what is a PlainKeyed table?

src/storage/kv/resource.rs
```diff
@@ -0,0 +1,146 @@
...
+        StorageScheme::RedisKv => {
+            let key_str = partition_key.to_string();
+
+            let drainer_query = model
+                .insert_drainer_query()
+                .map_err(|e| kv_write_error::<M::Error>(KvWriteError::Backend(e)))?;
+
+            let reply = kv_wrapper::<(), M>(
+                store,
+                KvOperation::HSetNx(M::ENTITY_TYPE, &model, drainer_query),
```
hrithikesh026:
```suggestion
            let diesel_model_after_change = M::into_domain(model);
            let reply = kv_wrapper::<(), M>(
                store,
                KvOperation::HSetNx(M::ENTITY_TYPE, diesel_model_after_change, drainer_query),
```

src/storage/kv/resource.rs
```diff
@@ -0,0 +1,146 @@
...
+                partition_key,
+            )
+            .await
+            .map_err(|e| {
+                kv_write_error::<M::Error>(KvWriteError::Backend(
+                    e.to_redis_failed_response(&key_str),
+                ))
```
hrithikesh026:
```suggestion
                Ok(HsetnxReply::KeySet) => Ok(diesel_model_after_change),
```

src/storage/kv/impls/fingerprint.rs
```diff
@@ -0,0 +1,99 @@
...
+            .await?;
+        Ok(output)
+    }
+}
+
+impl KvFindOptional for FingerprintTableNew {
+    async fn storage_find_optional(
+        store: &Storage,
+        pk: &PartitionKey<'_>,
+    ) -> Result<Option<Fingerprint>, crate::error::ContainerError<FingerprintDBError>> {
```
hrithikesh026: During DB failover, we need to start reading from READ replica right.???
```suggestion
        let mut conn = store.route_conn().await?;
```

src/storage/kv/impls/fingerprint.rs
```diff
@@ -0,0 +1,99 @@
...
+        store: &Storage,
+    ) -> Result<Fingerprint, crate::error::ContainerError<FingerprintDBError>> {
+        let mut conn = store.get_conn().await?;
+        let output: Fingerprint = diesel::insert_into(Fingerprint::table())
+            .values(self)
+            .get_result(&mut conn)
+            .await?;
+        Ok(output)
+    }
+}
```
hrithikesh026: It would make more sense to have this implementation on Fingerprint.
```suggestion
impl KvFindOptional for Fingerprint {
```

su-shivanshmathur: Got it

src/storage/kv/resource.rs
```diff
@@ -0,0 +1,146 @@
...
+pub(crate) trait StorageResource:
+    serde::Serialize
+    + serde::de::DeserializeOwned
+    + std::fmt::Debug
+    + KvStorePartition
+    + EntityType
+    + Sync
+    + Send
+{
+    type Domain;
```
hrithikesh026:
```suggestion
    type DieselModel: KvFindOptional;
```

src/storage/kv/resource.rs
```diff
@@ -0,0 +1,146 @@
...
+    fn set_storage_scheme(&mut self, scheme: StorageScheme);
+    fn insert_drainer_query(&self) -> error_stack::Result<SerializableQuery, KvError>;
+    async fn storage_insert(
+        self,
+        store: &Storage,
+    ) -> Result<Self::Domain, ContainerError<Self::Error>>;
+}
+
+/// DB find op for KV-routed tables. `StorageResource` is a super-trait bound.
+pub(crate) trait KvFindOptional: StorageResource {
```
hrithikesh026:
```suggestion
pub(crate) trait KvFindOptional {
```

src/storage/kv/resource.rs
```diff
@@ -0,0 +1,146 @@
...
+pub(crate) enum KvWriteError {
+    Duplicate,
+    Backend(Report<KvError>),
+}
+
+fn kv_write_error<E: From<KvWriteError> + error_stack::Context>(
+    e: KvWriteError,
+) -> ContainerError<E> {
+    ContainerError::from(E::from(e))
+}
```
hrithikesh026: can be an impl fn instead of hanging fn

src/storage/kv/resource.rs
```diff
@@ -0,0 +1,146 @@
...
+
+/// Marker for plain-keyed tables (Redis HGet/HSetNx).
+pub(crate) trait PlainKeyed {}
+
+#[derive(Debug)]
+pub(crate) enum KvWriteError {
+    Duplicate,
+    Backend(Report<KvError>),
+}
```
hrithikesh026: variants in this error enum seems like a subset of 'enum KvError'. Is this really needed. Isn't KvError enough?

src/storage/storage_v2/types.rs
```diff
@@ -57,6 +62,20 @@ impl From<VaultInner> for Vault {
...
+    }
+}
+
+impl From<VaultNew> for Vault {
+    fn from(value: VaultNew) -> Self {
+        Self {
+            vault_id: value.vault_id,
+            entity_id: value.entity_id,
+            data: value.encrypted_data.into(),
+            created_at: time::PrimitiveDateTime::MIN,
```
hrithikesh026:
```suggestion
            created_at: value.created_at
```

hrithikesh026: Also I just checked. created_at column has default constraint in DB. that should not be the case. Application should always write the value instead of letting DB calculate it.

src/storage/scheme.rs
```diff
@@ -0,0 +1,50 @@
...
+impl ToSql<Text, Pg> for StorageScheme {
+    fn to_sql<'b>(&'b self, out: &mut serialize::Output<'b, '_, Pg>) -> serialize::Result {
+        <str as ToSql<Text, Pg>>::to_sql(self.as_ref(), out)
+    }
+}
+
+impl FromSql<Text, Pg> for StorageScheme {
+    fn from_sql(bytes: <Pg as Backend>::RawValue<'_>) -> deserialize::Result<Self> {
+        let s = <String as FromSql<Text, Pg>>::from_sql(bytes)?;
+        s.parse::<Self>()
```
hrithikesh026, line 47: lets use from_str

src/storage/scheme.rs
```diff
@@ -0,0 +1,50 @@
...
+#[strum(serialize_all = "snake_case")]
+pub enum StorageScheme {
+    #[default]
+    PostgresOnly,
+    RedisKv,
+}
+
+impl ToSql<Text, Pg> for StorageScheme {
+    fn to_sql<'b>(&'b self, out: &mut serialize::Output<'b, '_, Pg>) -> serialize::Result {
+        <str as ToSql<Text, Pg>>::to_sql(self.as_ref(), out)
```
hrithikesh026, line 40: call to_string() on self

src/storage/types.rs
```diff
@@ -227,11 +234,12 @@ impl std::ops::Deref for CardNumber {
     }
 }
 
-#[derive(Debug, Insertable)]
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize, Insertable)]
```
hrithikesh026: no need for serde::Serialize. We will never insert New record into redis.

```suggestion
#[derive(Debug, Clone, serde::Deserialize, Insertable)]
```

src/config.rs
```diff
@@ -363,6 +386,95 @@ impl GlobalConfig {
...
+    /// TTL (seconds) for KV keys in Redis. Must exceed max drainer replay lag.
+    #[serde(default = "default_ttl_for_kv")]
+    pub ttl_for_kv: u32,
+    /// KV master switch. Overridden by runtime config when enabled.
+    #[serde(default)]
+    pub(crate) kv_state: crate::storage::kv::KvState,
+}
+
+#[cfg(feature = "kv")]
+fn default_drainer_stream_suffix() -> String {
```
hrithikesh026, line 433: I think we should limit kv state to be inferred from the runtime config endpoint alone no???

migrations/2026-06-30-000000_add_updated_by_to_kv_tables/up.sql
```diff
@@ -0,0 +1,21 @@
+-- Add `updated_by` to the KV-participating tables.
+--
+-- Tracks which storage backend wrote a row (`postgres_only` or `redis_kv`)
+-- so the drainer and read paths can reason about provenance.
+-- `schema.rs` already declares these columns; this migration brings existing
+-- databases in line (it was missing from the original KV PR #177).
```
SanchithHegde: Can we remove these 2 lines / this comment?

migrations/2026-06-30-000000_add_updated_by_to_kv_tables/up.sql
```diff
@@ -0,0 +1,21 @@
...
+--
+-- `IF NOT EXISTS` guards against drift on databases where the column was added
+-- manually before this migration landed.
+
+ALTER TABLE fingerprint
+    ADD COLUMN IF NOT EXISTS updated_by VARCHAR;
```
SanchithHegde: Are we not specifying the size? `VARCHAR(32)`, `VARCHAR(64)`, etc.?

src/error/kv.rs
```diff
@@ -0,0 +1,72 @@
+use hyperswitch_redis_interface::errors::RedisError;
+
+#[derive(Debug, thiserror::Error)]
+pub enum KvError {
+    #[error("DuplicateValue: {entity} already exists {key:?}")]
+    DuplicateValue {
+        entity: &'static str,
+        key: Option<String>,
+    },
```
SanchithHegde: Why is the `key` field optional? And are we raising this enum variant anywhere?

src/error/kv.rs
```diff
@@ -0,0 +1,72 @@
...
+    #[error("KV backend error")]
+    Backend,
+    #[error("ValueNotFound: {0}")]
+    ValueNotFound(String),
```
SanchithHegde: Nit: Would prefer using human readable error messages, rather than using variant names in error messages. (You can choose to keep or drop the backticks, I don't mind.)


```suggestion
    #[error("Duplicate value: {entity} `{key:?}` already exists")]
    DuplicateValue {
        entity: &'static str,
        key: Option<String>,
    },
    #[error("KV backend error")]
    Backend,
    #[error("Value not found: {0}")]
    ValueNotFound(String),
```

src/error/kv.rs
```diff
@@ -0,0 +1,72 @@
...
+            | RedisError::StreamTrimFailed
+            | RedisError::StreamAcknowledgeFailed
+            | RedisError::StreamEmptyOrNotAvailable
+            | RedisError::ConsumerGroupCreateFailed
+            | RedisError::ConsumerGroupDestroyFailed
+            | RedisError::ConsumerGroupRemoveConsumerFailed
+            | RedisError::ConsumerGroupSetIdFailed
+            | RedisError::ConsumerGroupClaimFailed
+            | RedisError::JsonSerializationFailed
+            | RedisError::JsonDeserializationFailed
```
SanchithHegde: Shouldn't we map these two (or the first one at least) to `SerializationFailed` variant? Or if the `SerializationFailed` variant is being used to indicate drainer SQL query generation failures alone, can we then name that enum variant accordingly?

src/storage/kv/impls.rs
SanchithHegde, line 1: Please refrain from using `mod.rs` files throughout the codebase. We're following Rust 2018 edition's `foo.rs` convention of naming module files, instead of `mod.rs` files.
See: https://doc.rust-lang.org/stable/edition-guide/rust-2018/path-changes.html#no-more-modrs

We can also consider enabling a clippy lint for the same, similar to how we have it in the hyperswitch repo:
https://github.com/juspay/hyperswitch/blob/230b2b4a251d6d190fe5dbf5f65cd839f786a1c4/Cargo.toml#L29

src/storage/kv/metrics.rs
```diff
@@ -0,0 +1,10 @@
+//! KV operation counter metrics.
+
+use crate::{counter_metric, global_meter};
+
+global_meter!(pub(crate) KV_METER, "card_vault_kv");
+
+counter_metric!(pub(crate) KV_OPERATION_SUCCESSFUL, KV_METER);
+counter_metric!(pub(crate) KV_OPERATION_FAILED, KV_METER);
+counter_metric!(pub(crate) KV_PUSHED_TO_DRAINER, KV_METER);
+counter_metric!(pub(crate) KV_FAILED_TO_PUSH_TO_DRAINER, KV_METER);
```
SanchithHegde: Nits:

1. Can consider keeping success / failure as attributes, with `result` = `success` / `failure` being the attribute.
2. Can consider specifying a lowercase name and optional unit / description to keep it consistent with the existing metrics.


```suggestion
counter_metric!(
    pub(crate) KV_OPERATIONS, KV_METER,
    name: "kv.operations",
    description: "Count of KV operations",
    unit: "1",
);
counter_metric!(
    pub(crate) KV_DRAINER_ENQUEUES, KV_METER,
    name: "kv.drainer.enqueues",
    description: "Count of KV drainer enqueue attempts",
    unit: "1",
);
```

Let me know what you think.

src/storage/kv.rs
```diff
@@ -0,0 +1,24 @@
...
+pub(crate) mod scheme;
+pub(crate) mod serializable_query;
+pub(crate) mod wrapper;
+
+#[cfg(feature = "kv")]
+pub(crate) use partition_key::PartitionKey;
+#[cfg(feature = "kv")]
+pub(crate) use resource::{find_optional_plain_resource, insert_plain_resource};
+pub(crate) use scheme::KvState;
+pub(crate) use wrapper::{KvStoreContext, RedisConnInterface};
```
SanchithHegde: Nit, optional change: Can use `self::` to indicate that these are being pulled from submodules.


```suggestion
#[cfg(feature = "kv")]
pub(crate) use self::{entity::EntityType, partition_key::PartitionKey, resource::{FindResourceBy, InsertResourceParams, find_optional_resource_by_id, insert_resource}};
pub(crate) use self::{scheme::KvState, wrapper::{KvStoreContext, RedisConnInterface}};
```

Please run formatter as required.

src/storage/kv/request_id.rs
```diff
@@ -0,0 +1,14 @@
+//! Request-id propagation via a Tokio task-local.
+//! Set per-request by the `scope_request_id` middleware in `app.rs`.
+//! Task-locals do not cross `tokio::spawn`.
+
+use tokio::task_local;
+
+task_local! {
+    pub(crate) static REQUEST_ID: String;
+}
```
SanchithHegde: Why do we need `tokio` task-local data for request ID? What other options do we have to store and retrieve the request ID for every request?

src/storage/scheme.rs
su-shivanshmathur: This is done by using `as_ref()`

src/storage/scheme.rs
su-shivanshmathur: tried to use `.parse()` for that, i hope its fine ?

src/config.rs
su-shivanshmathur: Added

hrithikesh026 (dismissed):

src/storage/kv/scheme.rs
```diff
@@ -0,0 +1,104 @@
...
+pub(crate) enum KvState {
+    #[default]
+    Disabled,
+    /// Write-through Redis; drainer replays to Postgres.
+    Enabled,
+    /// Insert to Postgres only; reads prefer Redis.
+    SoftKill,
+}
+
+impl<'de> serde::Deserialize<'de> for KvState {
```
SanchithHegde: Why do we need a custom `Deserialize` implementation, can't we keep it simple and just derive `Deserialize` with `serde(rename_all = "snake_case")`?

src/storage/kv/scheme.rs
```diff
@@ -0,0 +1,104 @@
...
+}
+
+impl std::fmt::Display for Op {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::Insert => f.write_str("insert"),
+            Self::Find => f.write_str("find"),
+        }
+    }
+}
```
SanchithHegde: - Nit: We can also derive `Copy`.
- Nit: We can derive `strum::Display` (with snake-case renaming) instead of the manual `Display` impl.

Both are optional changes.

src/storage/kv/scheme.rs
```diff
@@ -0,0 +1,104 @@
...
+        KvState::SoftKill => {
+            let scheme = match operation {
+                Op::Insert => StorageScheme::PostgresOnly,
+                Op::Find => StorageScheme::RedisKv,
+            };
+            debug!(%scheme, %operation, "soft-kill routing");
+            scheme
+        }
+    }
+}
```
SanchithHegde, line 50: What about update operation, are we not accounting for it here?

src/storage/kv/serializable_query.rs
```diff
@@ -0,0 +1,210 @@
...
+
+impl diesel::pg::PgMetadataLookup for KvPgMetadataLookup {
+    fn lookup_type(
+        &mut self,
+        _type_name: &str,
+        _schema: Option<&str>,
+    ) -> diesel::pg::PgTypeMetadata {
+        diesel::pg::PgTypeMetadata::from_result(Ok((FAKE_OID, FAKE_OID)))
+    }
+}
```
SanchithHegde: Why do we need this? The implementation in the Hyperswitch repo doesn't have this?

src/storage/storage_v2/db.rs
```diff
@@ -39,7 +39,6 @@ impl VaultInterface for Storage {
 
         logger::info!("performing retrieve operation on vault data");
 
-        // A missing row surfaces (via `?`) as `VaultDBError::NotFoundError`.
```
SanchithHegde, line 42: Can we add this comment back / Can we revert this change?

src/storage/db.rs
```diff
@@ -106,7 +111,6 @@ impl super::LockerInterface for Storage {
     ) -> Result<types::Locker, ContainerError<Self::Error>> {
         let mut conn = self.get_conn().await?;
 
-        // A missing row surfaces (via `?`) as `VaultDBError::NotFoundError`.
```
SanchithHegde, line 109: Can we add this comment back / Can we revert this change?

src/storage/redis.rs
SanchithHegde, line 1: Can we revert the changes to this file?

src/storage/db.rs
```diff
@@ -171,6 +175,7 @@ impl super::HashInterface for Storage {
         &self,
         data_hash: &[u8],
     ) -> Result<Option<types::HashTable>, ContainerError<Self::Error>> {
+        // Non-PK lookup — always Postgres.
```
SanchithHegde: Why did we add this comment?

src/storage/db.rs
```diff
@@ -192,6 +197,7 @@ impl super::HashInterface for Storage {
             .values(types::HashTableNew {
                 hash_id: utils::generate_uuid(),
                 data_hash,
+                updated_by: StorageScheme::PostgresOnly,
```
SanchithHegde, line 199: Why are we hardcoding `PostgresOnly` here?

src/storage/db.rs
```diff
@@ -220,6 +226,7 @@ impl super::TestInterface for Storage {
                         .values(types::HashTableNew {
                             hash_id: "test".to_string(),
                             data_hash: b"0".to_vec(),
+                            updated_by: StorageScheme::PostgresOnly,
```
SanchithHegde, line 229: Same here, why are we hardcoding the value here?

src/storage/types.rs
SanchithHegde, line 144: Why are we hardcoding this value?

src/storage/types.rs
```diff
@@ -227,11 +234,13 @@ impl std::ops::Deref for CardNumber {
     }
 }
 
-#[derive(Debug, Insertable)]
+// Not serialized to Redis — the Queryable model (`Fingerprint` above) is the stored value.
+#[derive(Debug, Clone, serde::Deserialize, Insertable)]
```
SanchithHegde: Nit: Do we need the `Deserialize` derive either?

src/storage/kv/resource.rs
```diff
@@ -0,0 +1,171 @@
...
+    + Send
+    + Sized
+{
+    type Error: error_stack::Context + Send + Sync + 'static + for<'a> From<&'a KvError>;
+
+    fn set_storage_scheme(&mut self, scheme: StorageScheme);
+
+    /// Drainer INSERT — built from the `Insertable` `New` projection (the model is not
+    /// `Insertable`). Implementations rebuild the `New` struct from the model's fields.
+    fn insert_drainer_query(&self) -> error_stack::Result<SerializableQuery, KvError>;
```
SanchithHegde: Can we name this slightly differently? The current name makes me think that the query would be _inserted_ somewhere, while the implementation only generates the `INSERT` query and doesn't write it anywhere.

src/storage/kv/impls/fingerprint.rs
```diff
@@ -0,0 +1,87 @@
...
+            .get_result(&mut conn)
+            .await?)
+    }
+
+    async fn storage_find_optional(
+        store: &Storage,
+        pk: &PartitionKey<'_>,
+    ) -> Result<Option<Self>, ContainerError<FingerprintDBError>> {
+        let PartitionKey::Fingerprint { fingerprint_hash } = pk;
+        let mut conn = store.route_conn().await?;
```
SanchithHegde, line 65: Why are we using `route_conn()` here? Shouldn't we be always using the write replica for this table?

src/storage/kv/wrapper.rs
```diff
@@ -0,0 +1,250 @@
...
+    op: KvOperation<'a, S>,
+    partition_key: PartitionKey<'a>,
+) -> error_stack::Result<KvResult<T>, RedisError>
+where
+    T: de::DeserializeOwned,
+    S: serde::Serialize + Debug + KvStorePartition + Sync,
+{
+    let redis_conn = store.get_redis_conn()?;
+
+    let key = format!("{partition_key}");
```
SanchithHegde: Nit: Can also do:


```suggestion
    let key = partition_key.to_string();
```

src/storage/kv/wrapper.rs
```diff
@@ -0,0 +1,250 @@
...
+
+    let type_name = std::any::type_name::<T>();
+    let operation = op.to_string();
+
+    let ttl = store.ttl_for_kv();
+
+    let result = async {
+        match op {
+            KvOperation::HSetNx(field, value, query) => {
+                debug!(kv_operation = %operation, ?value);
```
SanchithHegde: Could it be a concern (from the perspective of sensitive information) that we're logging the value we're inserting?

src/storage/kv/impls/fingerprint.rs
```diff
@@ -0,0 +1,87 @@
...
+            .values(new)
+            .get_result(&mut conn)
+            .await?)
+    }
+
+    async fn storage_find_optional(
+        store: &Storage,
+        pk: &PartitionKey<'_>,
+    ) -> Result<Option<Self>, ContainerError<FingerprintDBError>> {
+        let PartitionKey::Fingerprint { fingerprint_hash } = pk;
```
SanchithHegde, line 63: As of now, there's only one enum variant, so this holds. But as more variants are added to `PartitionKey`, this assignment won't be as simple?

src/storage/db.rs
```diff
@@ -261,33 +268,79 @@ impl super::FingerprintInterface for Storage {
...
-        let mut conn = self.get_conn().await?;
+        #[cfg(feature = "kv")]
+        {
+            let fp_bytes = fingerprint_hash.peek().clone();
+            let partition_key = super::kv::PartitionKey::Fingerprint {
+                fingerprint_hash: &fp_bytes,
+            };
+
+            // Store & return the Queryable model, not the New struct.
+            return super::kv::find_optional_resource_by_id::<types::Fingerprint>(
```
SanchithHegde: W.r.t. the comment, what are we "storing" here?

src/storage/db.rs
```diff
@@ -261,33 +268,79 @@ impl super::FingerprintInterface for Storage {
...
+        {
+            // `id: 0` — serial unknown at KV-insert time, assigned by drainer on replay.
+            let model = types::Fingerprint {
+                id: 0,
+                fingerprint_hash: fingerprint_hash.clone(),
                 fingerprint_id,
-            })
-            .get_result(&mut conn)
-            .await?;
+                updated_by: StorageScheme::PostgresOnly,
```
SanchithHegde, line 313: Why is this hardcoded?

src/config.rs
```diff
@@ -363,6 +386,93 @@ impl GlobalConfig {
...
+    #[serde(default = "default_drainer_stream_suffix")]
+    pub drainer_stream_suffix: String,
+    /// Drainer shard count. Must be `> 0` (validated in [`KvConfig::validate`]).
+    #[serde(default = "default_drainer_num_partitions")]
+    pub drainer_num_partitions: u8,
+    /// TTL (seconds) for KV keys in Redis. Must exceed max drainer replay lag.
+    #[serde(default = "default_ttl_for_kv")]
+    pub ttl_for_kv: u32,
+    // KV enablement is sourced solely from the runtime-config endpoint (`locker.enable_kv`),
+    // fail-closed to `Disabled`. No TOML fallback — single source of truth.
```
SanchithHegde: Should we consider removing this comment?

src/custom_extractors.rs
SanchithHegde, line 1: Nit: Can we revert changes to this file?

src/runtime_config.rs
SanchithHegde, line 1: Why was this file modified?

SanchithHegde (commented):
By the way, did you test this out end-to-end by running the drainer as well?

su-shivanshmathur:
> By the way, did you test this out end-to-end by running the drainer as well?

Yes @SanchithHegde

src/storage/kv/scheme.rs
su-shivanshmathur: A plain derive only accepts the bare string. The runtime-config endpoint sends `{"kv_state": "..."}` as an object, so we need the custom impl to handle both forms.

src/runtime_config.rs
su-shivanshmathur: Can't revert this. We added the `RuntimeConfigItem` trait and a generic `get::<T>()` method so both `ReplicaRouting` and `KvState` can read their configs from the runtime endpoint.

src/storage/kv/serializable_query.rs
su-shivanshmathur: Hyperswitch has a live database connection to resolve PostgreSQL type metadata. We build queries offline without one, so we need this stub. The dummy OIDs are never used — the drainer resolves real ones when replaying with a live connection

src/error/kv.rs
su-shivanshmathur: Done

src/storage/kv/request_id.rs
su-shivanshmathur: The request ID is set by middleware but needed deep inside `push_to_drainer_stream`. Passing it through every function would be an issue right ?
Task-local would set once and read anywhere in the same async task should be fine right ?

src/storage/kv/scheme.rs
su-shivanshmathur: Fingerprint only has insert/find. So Update will be added when vault table

src/storage/kv/impls.rs
su-shivanshmathur: Sure, Have added clippy lint under cargo.toml for the same

src/storage/kv/impls/fingerprint.rs
```diff
@@ -0,0 +1,89 @@
...
+            entity::EntityType,
+            partition_key::{KvStorePartition, PartitionKey},
+            resource::KvResource,
+            serializable_query::{SerializableQuery, generate_insert_query},
+        },
+        types::{Fingerprint, FingerprintTableNew},
+    },
+};
+
+// Required by `generate_insert_query<N: Insertable + EntityType>` — the drainer builder.
```
SanchithHegde: Nit: This would be required for generating update query as well, when we do add support for it.

src/storage/kv/impls/fingerprint.rs
```diff
@@ -0,0 +1,89 @@
...
+    fn set_storage_scheme(&mut self, scheme: StorageScheme) {
+        self.updated_by = scheme;
+    }
+
+    fn generate_insert_drainer_query(&self) -> error_stack::Result<SerializableQuery, KvError> {
+        let new = FingerprintTableNew {
+            fingerprint_hash: self.fingerprint_hash.clone(),
+            fingerprint_id: self.fingerprint_id.clone(),
+            updated_by: self.updated_by,
+        };
```
SanchithHegde: Nit: This could have been a `From` impl.

src/storage/kv/impls/fingerprint.rs
```diff
@@ -0,0 +1,89 @@
...
+
+impl From<&KvError> for FingerprintDBError {
+    fn from(e: &KvError) -> Self {
+        match e {
+            KvError::DuplicateValue { .. } => Self::Duplicate,
+            KvError::ValueNotFound(_) => Self::DBFilterError,
+            KvError::Backend | KvError::SerializationFailed => Self::UnknownError,
+        }
+    }
+}
```
SanchithHegde, line 81: We would lose the context of the original error, would it be possible to call `change_context()` here?

src/storage/kv/request_id.rs
SanchithHegde: What would happen if another task is spawned? Would that child task be able to access the task local value from the parent task, or no?

src/storage/kv/resource.rs
```diff
@@ -0,0 +1,169 @@
+//! Generic KV resource trait, key-shape locators, and CRUD helpers.
+//!
+//! Stores and returns the Diesel Queryable model (not the `New` projection). Mirrors
+//! hyperswitch's `insert_resource` / `find_optional_resource_by_id`.
```
SanchithHegde: Nit: Do we need the "mirror hyperswitch's behvior" comment in doc comments though?

src/storage/kv/resource.rs
```diff
@@ -0,0 +1,169 @@
...
+use crate::{
+    error::{
+        ContainerError,
+        kv::{KvError, RedisErrorExt},
+    },
+    storage::Storage,
+};
+
+/// A KV-routed table's Diesel Queryable model: stored in Redis, read back, returned to
+/// callers. Analog of hyperswitch `StorageModel<D>`.
```
SanchithHegde: Nit: Can remove the "Analog of hyperswitch `StorageModel<D>`" part.

src/storage/kv/resource.rs
```diff
@@ -0,0 +1,169 @@
...
+    ) -> Result<Option<Self>, ContainerError<Self::Error>>;
+}
+
+pub(crate) struct InsertResourceParams<'a> {
+    pub partition_key: PartitionKey<'a>,
+    pub identifier: &'static str,
+}
+
+/// Locator for a find. `Id` = plain-keyed (single HGet/HSetNx field).
+/// SEAM: add `LookupId(String)` for reverse-lookup tables with their first consumer.
```
SanchithHegde: SEAM?

src/storage/kv/resource.rs
```diff
@@ -0,0 +1,169 @@
...
+            let key_str = key.to_string();
+            let result =
+                kv_wrapper::<M, M>(store, KvOperation::<M>::HGet(field), key.clone()).await;
+
+            match result {
+                Ok(KvResult::HGet(v)) => Ok(Some(v)),
+                Err(e) if matches!(e.current_context(), RedisError::NotFound) => {
+                    // Redis miss → fall back to Postgres (mirrors hyperswitch's
+                    // `try_redis_get_else_try_database_get`). In SoftKill this means
+                    // the key was never written to Redis, so we read from DB.
```
SanchithHegde: Nit: Do we need the "mirrors hyperswitch's behavior" part?

src/storage/kv/scheme.rs
```diff
@@ -0,0 +1,100 @@
...
+
+        deserializer.deserialize_any(KvStateVisitor)
+    }
+}
+
+/// Operation type used by `decide_storage_scheme`.
+///
+/// `Update` is not yet included — fingerprint only has insert/find. When vault
+/// migrates, `Op::Update` will be added with a Redis probe in
+/// SoftKill mode, matching hyperswitch's behavior.
```
SanchithHegde: Nit: Do we need this in the doc comment though?

src/storage/kv/serializable_query.rs
SanchithHegde: Drainer does not do any metadata lookups though?

https://github.com/juspay/hyperswitch/blob/8c2448303fa9b93cebeb6ef3712b8287dfab9259/crates/diesel_models/src/kv.rs#L121-L151

src/storage/db.rs
```diff
@@ -261,33 +270,81 @@ impl super::FingerprintInterface for Storage {
...
+            let partition_key = super::kv::PartitionKey::Fingerprint {
+                fingerprint_hash: fingerprint_hash.peek().as_slice(),
+            };
+
+            return super::kv::insert_resource::<types::Fingerprint>(
+                self,
+                model,
+                super::kv::InsertResourceParams {
+                    partition_key,
+                    identifier: types::Fingerprint::ENTITY_TYPE,
```
SanchithHegde: If we're passing entity name here, should we consider renaming this field?

src/storage/db.rs
```diff
@@ -302,9 +359,7 @@ impl super::EntityInterface for Storage {
         // Reads are routed to the read replica when enabled (#171).
         let mut conn = self.route_conn().await?;
 
-        // A missing row surfaces as `EntityDBError::NotFoundError` (see the `From<diesel>`
-        // classifier), which `find_or_create_entity` in the key manager checks via
-        // `is_not_found()`.
+        // Missing row → `EntityDBError::NotFoundError`, checked by `find_or_create_entity`.
```
SanchithHegde: Nit: Would prefer this change being reverted.

src/storage/kv/impls/fingerprint.rs
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    async fn storage_find_optional(
+        store: &Storage,
+        pk: &PartitionKey<'_>,
+    ) -> Result<Option<Self>, ContainerError<FingerprintDBError>> {
+        let PartitionKey::Fingerprint { fingerprint_hash } = pk;
+        // Use the primary conn — the drainer writes to the primary, and this PG
+        // fallback fires on a Redis miss where the row may have just been replayed.
+        let mut conn = store.get_conn().await?;
```
SanchithHegde: Guess we'll have to call `route_conn()` here.

src/storage/kv/scheme.rs
SanchithHegde: Can we define a separate struct maybe with all configs, and the config names being the keys / field names of the struct?

src/runtime_config.rs
```diff
@@ -70,8 +77,9 @@ impl RuntimeConfigManager {
             #[cfg_attr(not(feature = "caching"), allow(unused_variables))]
             RuntimeConfig::Enabled {
                 endpoint,
-                ttl_seconds,
+                ttl_seconds: _,
```
SanchithHegde: We can remove this field if this is not required.

src/runtime_config.rs
```diff
@@ -109,51 +118,119 @@ impl RuntimeConfigManager {
...
-        let raw = match Self::fetch(endpoint_url, api_key, client, key).await {
-            Ok(val) => val,
-            Err(error) => {
-                crate::logger::error!(?error, key, "Failed to fetch runtime config from endpoint");
-                return None;
+        #[cfg(not(feature = "caching"))]
+        {
+            let _ = key;
+            let _ = &self.state;
+        }
```
SanchithHegde: Nit: Why do we need this block?

src/runtime_config.rs
```diff
@@ -109,51 +118,119 @@ impl RuntimeConfigManager {
...
+            } => Duration::from_secs(*refresh_interval_seconds),
+        };
+
+        let manager = Arc::clone(self);
+        crate::logger::info!(
+            refresh_interval_secs = refresh_interval.as_secs(),
+            "Spawning runtime config prefetch task"
+        );
+
+        Some(tokio::spawn(async move {
```
SanchithHegde: Can you add `.in_current_span()`?

src/storage/kv/impls/fingerprint.rs
su-shivanshmathur: We're not losing the context here actually, as this `From` impl just maps `KvError` to the right `FingerprintDBError` variant, it can't do `change_context` since it returns `Self` and not a report here

src/storage/kv/serializable_query.rs
```diff
@@ -0,0 +1,212 @@
...
```
SanchithHegde, line 96: As suggested [before](https://github.com/juspay/hyperswitch-card-vault/pull/177#discussion_r3523306137), can we remove this and use a live connection to obtain the type metadata? Basically, can we replicate / follow what Hyperswitch does?

https://github.com/juspay/hyperswitch/blob/0809d84ba8df1dfa732ff3710c6979dcd5e50689/crates/diesel_models/src/kv.rs#L70-L119

src/storage/kv/wrapper.rs
```diff
@@ -0,0 +1,246 @@
...
+
+use super::{
+    metrics,
+    partition_key::{KvStorePartition, PartitionKey},
+    serializable_query::SerializableQuery,
+};
+use crate::logger;
+
+/// Drainer-entry `request_id`: log-only, empty (not threaded), kept for wire-format parity.
+const REQUEST_ID: &str = "VAULT_CONSTANT_REQUEST_ID";
```
SanchithHegde, line 20: I realized we could also generate a value when inserting the drainer query, and ensuring that we log the generated value on the locker side should still help us with debugging/correlation. Optional change.

src/storage/db.rs
SanchithHegde: I meant to say renaming from `identifier` to something like `entity_type`, not renaming it to `field`.

src/storage/kv/resource.rs
```diff
@@ -0,0 +1,168 @@
...
+    async fn storage_find_optional(
+        store: &Storage,
+        pk: &PartitionKey<'_>,
+    ) -> Result<Option<Self>, ContainerError<Self::Error>>;
+}
+
+pub(crate) struct InsertResourceParams<'a> {
+    pub partition_key: PartitionKey<'a>,
+    /// Redis hash field the value is stored under.
+    pub field: &'static str,
```
SanchithHegde: As mentioned earlier, can we rename this as `entity_type` maybe, since we are passing the `ENTITY_TYPE` constant when constructing this struct?

src/storage/kv/impls/fingerprint.rs
```diff
@@ -0,0 +1,91 @@
...
+
+impl From<&Fingerprint> for FingerprintTableNew {
+    fn from(fingerprint: &Fingerprint) -> Self {
+        Self {
+            fingerprint_hash: fingerprint.fingerprint_hash.clone(),
+            fingerprint_id: fingerprint.fingerprint_id.clone(),
+            updated_by: fingerprint.updated_by,
+        }
+    }
+}
```
SanchithHegde: Nit: Can we consider moving this impl to where the types are defined?

src/runtime_config.rs
```diff
@@ -86,74 +82,112 @@ impl RuntimeConfigManager {
...
         })
     }
 
-    #[inline]
-    fn deserialize_config<T: serde::de::DeserializeOwned>(key: &str, raw: &str) -> Option<T> {
+    /// Deserialize the last-known-good config body into `T`. A disabled manager, no config fetched
+    /// yet, or an unparseable payload yields `T::default()` (fail-closed).
+    pub async fn get<T: serde::de::DeserializeOwned + Default>(&self) -> T {
+        let RuntimeConfigState::Enabled { cache, .. } = &self.state else {
+            return T::default();
```
SanchithHegde: Nit: Can we add a log line as to why we returned `T::default()` here?

src/storage.rs
```diff
@@ -20,34 +20,46 @@ use crate::{
...
     replica_pg_pool: Option<Arc<Pool<AsyncPgConnection>>>,
     runtime_config_manager: Arc<crate::runtime_config::RuntimeConfigManager>,
+    #[cfg(feature = "kv")]
+    redis: Option<redis_store::RedisStore>,
+    #[cfg(feature = "kv")]
+    kv_config: crate::config::KvConfig,
 }
 
+#[cfg(feature = "kv")]
+use crate::storage::redis as redis_store;
```
SanchithHegde: Nit: Would prefer imports being placed near the top of the file.

config/config.example.toml
```diff
@@ -113,6 +113,26 @@ unresponsive_check_interval = 2    # how often to check for unresponsiveness (se
 max_feed_count = 200               # max commands fed to the server at once
 max_failure_threshold_seconds = 5  # max seconds Redis may be unreachable before it's failed
 
+# KV (write-through Redis) configuration — global drainer settings (only used with the `kv` feature)
+[kv]
+drainer_stream_suffix = "DRAINER_STREAM"  # suffix for the drainer stream name
+drainer_num_partitions = 16        # number of shard partitions (must be > 0)
+ttl_for_kv = 900                   # TTL for KV entries in Redis (seconds)
+
+# KV enablement is controlled SOLELY at runtime via the `locker.enable_kv` runtime-config
```
SanchithHegde: Nit: Is `locker.enable_kv` the key we use for fetching all runtime configs, based on the recent changes in this PR? If not, can we update this comment suitably?

src/runtime_config.rs
hrithikesh026: Better to return None instead of relying on a blind default implementation.

src/storage/kv/resource.rs
hrithikesh026: For card vault use case. We decided partition_key and hash_key field will remain same. So this is wrong implementation. Have changed it.

src/storage/kv/wrapper.rs
hrithikesh026: Will take up in a future PR

src/storage/kv/serializable_query.rs
hrithikesh026: It is not directly possible to follow Hyperswitch’s implementation here because Hyperswitch uses async_bb8_diesel, which exposes conn.run(...) with access to the underlying sync PgConnection. That lets them pass the live connection itself as PgMetadataLookup during collect_binds.

This repo uses diesel-async with AsyncPgConnection. In this setup, the equivalent Postgres metadata lookup machinery exists inside diesel-async, but it is private and not exposed for external query serialization.

For the current fingerprint insert, all bind types are built-in Postgres types (bytea, text), so Diesel can resolve metadata statically without a live lookup.

SanchithHegde: Would [`AsyncConnectionWrapper`](https://docs.rs/diesel-async/latest/diesel_async/async_connection_wrapper/type.AsyncConnectionWrapper.html) help us here?

src/storage/kv/resource.rs
```diff
@@ -0,0 +1,170 @@
...
+    M::set_storage_scheme(&mut diesel_new, scheme);
+
+    match scheme {
+        StorageScheme::PostgresOnly => M::storage_insert(diesel_new, store).await,
+        StorageScheme::RedisKv => {
+            let drainer_query = M::generate_insert_drainer_query(&diesel_new)
+                .map_err(kv_backend_error::<M::Error>)?;
+
+            let key_str = partition_key.to_string();
+            // apply the changes in applciation
```
SanchithHegde: Nit: Typo.


```suggestion
            // apply the changes in application
```

src/storage/types.rs
```diff
@@ -180,14 +186,27 @@ pub struct HashTable {
...
+#[derive(Debug, Clone, Identifiable, Queryable, serde::Serialize, serde::Deserialize)]
 #[diesel(table_name = schema::fingerprint)]
 pub struct Fingerprint {
     pub id: i32,
     pub fingerprint_hash: Secret<Vec<u8>>,
     pub fingerprint_id: Secret<String>,
+    pub updated_by: StorageScheme,
+}
+
+impl From<FingerprintTableNew> for Fingerprint {
```
SanchithHegde, line 201: Nit: Is this `From` impl being used anywhere now?

src/runtime_config.rs
```diff
@@ -86,74 +82,112 @@ impl RuntimeConfigManager {
...
                     },
                 }
             }
         })
     }
 
-    #[inline]
-    fn deserialize_config<T: serde::de::DeserializeOwned>(key: &str, raw: &str) -> Option<T> {
+    /// Deserialize the last-known-good config body into `T`. A disabled manager, no config fetched
+    /// yet, or an unparseable payload yields `T::default()` (fail-closed).
```
SanchithHegde: Nit: Typo.

```suggestion
    /// yet, or an unparsable payload yields `T::default()` (fail-closed).
```

src/runtime_config.rs
```diff
@@ -86,74 +82,112 @@ impl RuntimeConfigManager {
...
+    pub async fn get<T: serde::de::DeserializeOwned + Default>(&self) -> Option<T> {
+        let RuntimeConfigState::Enabled { cache, .. } = &self.state else {
+            return None;
+        };
```
SanchithHegde, line 104: Should we consider adding a log line here as to why we're returning `None`, or do you think it's not required?

src/storage/types.rs
hrithikesh026: Yes. there is a new GAT in `KvResource` trait that requires this.

src/runtime_config.rs
hrithikesh026: Added more information on when this function returns None in the function documentation.

src/storage/kv/serializable_query.rs
hrithikesh026:
```
cargo check --features "release"
   Compiling hyperswitch_card_vault v0.1.3 (/Users/hrithikesh.vm/Desktop/OSS/hyperswitch-card-vault-2)
error[E0277]: the trait bound `AsyncPgConnection: PgMetadataLookup` is not satisfied
   --> src/storage/kv/serializable_query.rs:163:58
    |
163 |                 query.collect_binds(&mut bind_collector, conn, &Pg)?;
    |                                                          ^^^^ the trait `GetPgMetadataCache` is not implemented for `AsyncPgConnection`
    |
help: the trait `GetPgMetadataCache` is implemented for `PgConnection`
   --> /Users/hrithikesh.vm/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/diesel-2.2.12/src/pg/connection/mod.rs:273:1
    |
273 | impl GetPgMetadataCache for PgConnection {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = note: required for `AsyncPgConnection` to implement `PgMetadataLookup`
    = note: required for the cast from `&mut AsyncPgConnection` to `&mut (dyn PgMetadataLookup + 'static)`

error[E0277]: the trait bound `AsyncPgConnection: PgMetadataLookup` is not satisfied
   --> src/storage/kv/serializable_query.rs:163:58
    |
163 |                 query.collect_binds(&mut bind_collector, conn, &Pg)?;
    |                                                          ^^^^ the trait `LoadConnection` is not implemented for `AsyncPgConnection`
    |
help: the trait `LoadConnection<B>` is implemented for `PgConnection`
   --> /Users/hrithikesh.vm/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/diesel-2.2.12/src/pg/connection/mod.rs:244:1
    |
244 | / impl<B> LoadConnection<B> for PgConnection
245 | | where
246 | |     Self: self::private::PgLoadingMode<B>,
    | |__________________________________________^
    = note: required for `AsyncPgConnection` to implement `PgMetadataLookup`
    = note: required for the cast from `&mut AsyncPgConnection` to `&mut (dyn PgMetadataLookup + 'static)`

For more information about this error, try `rustc --explain E0277`.
error: could not compile `hyperswitch_card_vault` (lib) due to 2 previous errors
```

technically two trait bounds are not satisfied for `AsyncPgConnection` and `AsyncConnectionWrapper` 
those traits are `PgMetadataLookup` and `GetPgMetadataCache`. These traits are only available for `PgConnection`. Since these are postgres specific traits.

hrithikesh026: For the scope of this PR, I feel like we can ignore this for now since we don't use any custom DB types.

SanchithHegde (approved):

PR #178, opened 2026-06-30
chore(runtime): add support for optional static headers in runtime config requests

su-shivanshmathur:
## Summary

Operators can now attach extra static headers (e.g. `accept`) to every runtime config pull request via `[runtime_config.endpoint.headers]` in TOML. No-op when unconfigured.

## Behavior

- Headers omitted → empty map → `default_headers` is no-op → byte-identical to today
- Headers present → baked into the client once at startup, sent on every pull

### DIFF
--- a/src/runtime_config.rs
+++ b/src/runtime_config.rs
@@ -1,4 +1,4 @@
-use std::time::Duration;
+use std::{collections::HashMap, time::Duration};
 
 use error_stack::ResultExt;
 use hyperswitch_masking::{PeekInterface, Secret};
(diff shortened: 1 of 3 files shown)


src/runtime_config.rs
```diff
@@ -37,6 +37,27 @@ pub struct RuntimeConfigManager {
...
+                )),
+            )?;
+            let mut hval = reqwest::header::HeaderValue::from_str(value.peek()).change_context(
+                error::ConfigurationError::InvalidConfigurationValueError(format!(
+                    "invalid runtime_config header value for `{name}`"
+                )),
+            )?;
+            hval.set_sensitive(true);
+            map.insert(hname, hval);
+        }
```
hrithikesh026: lets do idiomatic .iter().map().collect()

su-shivanshmathur: Done

hrithikesh026 (approved):

ShankarSinghC (approved):

PR #180, opened 2026-07-14
feat(kv): add reverse lookup kv support

hrithikesh026:
## Summary
- Adds KV-backed storage support for reverse lookup records, including insert, find, and delete paths behind the `kv` feature.
- Introduces `ReverseLookup` as a KV resource with reverse lookup partitioning, serialization, drainer insert/delete queries, and KV error mapping.
- Cleans up reverse lookup records when deleting lockers from KV to avoid stale locker secondary lookup entries.
- Splits KV delete behavior for resources that delete only their primary entry versus resources that also own reverse lookup records.

## Changes
- Added `PartitionKey::ReverseLookup` and `src/storage/kv/impls/reverse_lookup.rs`.
- Added `delete_reverse_lookup` to `ReverseLookupInterface` with KV and DB implementations.
- Added `KvDeleteWithoutLookup` / `KvDeletableWithLookup` deletion traits and `delete_resource_by_id_with_reverse_lookup`.
- Marked Vault and ReverseLookup deletions as lookup-free, and Locker deletion as reverse-lookup-aware.
- Added reverse lookup duplicate error handling and serde derives for `ReverseLookup`.

## Testing
- Not run locally.


hrithikesh026:
Add delete support as well

hrithikesh026:
> Add delete support as well


Also delete reverse lookup when the parent record id deleted.

src/storage/db.rs
```diff
@@ -652,58 +653,114 @@ impl super::ReverseLookupInterface for Storage {
...
+            let output: types::ReverseLookup =
+                super::record_db_query::<<types::ReverseLookup as HasTable>::Table, _, _, _>(
+                    query.get_result(conn.get_mut()),
+                    operation,
+                    pool,
+                )
+                .await?;
+            Ok(output)
         }
-        .map_err(From::from)
```
SanchithHegde, line 683: Any specific reason to be dropping this error handling logic?

SanchithHegde (approved):

src/storage/db.rs
hrithikesh026: log_db_query can return `ReverseLookupDBError` directly through the generic. No need for getting `diesel::result::Error `out first and then converting it to `ReverseLookupDBError`.

PR #181, opened 2026-07-14
feat(kv): add vault kv support

hrithikesh026:
## Summary
- Adds KV-backed vault CRUD support behind the `kv` feature, including `VaultPrimaryKey`, vault partition keys, and `KvResource`/`KvUpdateResource`/`KvDeleteResource` implementations.
- Routes vault insert, find, update, and delete through the generic KV helpers when `kv` is enabled, while preserving the existing Postgres/Diesel paths when it is not.
- Extends the generic KV framework with update/delete support: `Op::Update`, `Op::Delete`, Redis `Hset`/`HDel` operations, result handling, and drainer serialization for update/delete queries.
- Adds locker delete support through the generic KV delete path.
- Makes vault storage models serializable/deserializable for Redis and adds conversion helpers for KV insert/update payloads.

## Testing
- `cargo fmt --check`
- `cargo check --features kv`


src/storage/kv/resource.rs
```diff
@@ -337,6 +379,85 @@ where
...
+
+    match scheme {
+        StorageScheme::PostgresOnly => M::storage_update(store, diesel_new, primary_key).await,
+        StorageScheme::RedisKv => {
+            let key = primary_key.get_partition_key();
+            let current = find_resource_by_id::<M>(store, primary_key.clone()).await?;
+            let update_query = M::generate_update_drainer_query(&diesel_new, &primary_key)
+                .map_err(kv_backend_error::<M::Error>)?;
+            let updated_model = M::apply_update(diesel_new, current);
+            let redis_value = serde_json::to_string(&updated_model).map_err(|err| {
```
hrithikesh026: no need for this. updated_model should implement serialize trait.

src/storage/kv/scheme.rs
```diff
@@ -40,7 +40,7 @@ pub(crate) fn decide_storage_scheme(state: KvState, operation: Op) -> StorageSch
         KvState::Enabled => StorageScheme::RedisKv,
         KvState::SoftKill => {
             let scheme = match operation {
-                Op::Insert => StorageScheme::PostgresOnly,
+                Op::Insert | Op::Update | Op::Delete => StorageScheme::PostgresOnly,
```
hrithikesh026: As long as the record in available cached in redis, update and delete should. go through KV.

Get the Key from the caller and find out if the value is available in redis through HGET with kv_wrapper and decide

src/storage/storage_v2/types.rs
```diff
@@ -10,7 +10,7 @@ use crate::{
     },
 };
 
-#[derive(Debug)]
+#[derive(Debug, serde::Serialize, serde::Deserialize)]
```
hrithikesh026: these serde derives are not needed for this.

src/storage/types.rs
```diff
@@ -105,7 +105,7 @@ pub struct Locker {
     pub updated_by: StorageScheme,
 }
 
-#[derive(Debug)]
+#[derive(Debug, serde::Serialize, serde::Deserialize)]
```
hrithikesh026: this serde derive is also not required is also not needed.

SanchithHegde (approved):

PR #182, opened 2026-07-14
feat(kv): add locker kv support

hrithikesh026:
Summary:
- Adds locker KV insert/read support using typed primary keys for partition key generation.
- Adds typed reverse lookup support for locker hash lookups via `LockerHashLookupKey` (`hash_id`, `merchant_id`, `customer_id`).
- Splits reverse-lookup behavior into `KvReverseLookupResource` so resources without secondary lookups, like fingerprint, do not need dummy reverse-lookup implementations.
- Adds `insert_resource_with_reverse_lookup` while keeping plain `insert_resource` for non-reverse-lookup KV resources.
- Wires `find_optional_by_hash_id_merchant_id_customer_id` through reverse lookup with Postgres fallback on reverse-lookup/Redis misses.
- Adds `ReverseLookup::get_partition_key` / `PartitionKey::CombinationKey` for stored reverse lookup partition keys.

Visual Overview:

```mermaid
classDiagram
    class KvResource {
        <<trait>>
        PrimaryKeyType: GetPartitionKey
        storage_insert()
        storage_find()
    }
    class KvReverseLookupResource {
        <<trait>>
        LookupKeyType: GetLookupKey
        get_reverse_lookup_key()
        storage_find_by_lookup()
    }
    class GetPartitionKey {
        <<trait>>
        get_partition_key()
    }
    class GetLookupKey {
        <<trait>>
        get_lookup_key()
    }
    class Locker
    class Fingerprint
    class LockerPrimaryKeyType
    class FingerprintPrimaryKey
    class LockerHashLookupKey

    KvReverseLookupResource --|> KvResource
    KvResource --> GetPartitionKey : PrimaryKeyType
    KvReverseLookupResource --> GetLookupKey : LookupKeyType
    Locker ..|> KvResource
    Locker ..|> KvReverseLookupResource
    Fingerprint ..|> KvResource
    LockerPrimaryKeyType ..|> GetPartitionKey
    FingerprintPrimaryKey ..|> GetPartitionKey
    LockerHashLookupKey ..|> GetLookupKey
```

Locker KV Insert With Reverse Lookup:

```mermaid
flowchart TD
    A[LockerInterface::insert_locker] --> B[Build PartitionKey::Locker]
    B --> C[insert_resource_with_reverse_lookup<Locker>]
    C --> D{Insert storage scheme}
    D -->|PostgresOnly| E[Locker::storage_insert]
    D -->|RedisKv| F[Generate drainer INSERT query]
    F --> G[Locker::get_reverse_lookup_key]
    G --> H[LockerHashLookupKey]
    H --> I[ReverseLookupKey lookup_id]
    I --> J[Insert reverse_lookup row]
    J --> K[HSetNx locker payload in Redis]
    K --> L[Push drainer stream entry]
    L --> M[Return locker model]
```

Locker Hash Lookup Read Path:

```mermaid
flowchart TD
    A[find_optional_by_hash_id_merchant_id_customer_id] --> B[LockerHashLookupKey hash_id + merchant_id + customer_id]
    B --> C[find_optional_resource_by_lookup_id<Locker>]
    C --> D{Find storage scheme}
    D -->|PostgresOnly| E[Locker::storage_find_by_lookup]
    D -->|RedisKv| F[Compute lookup_id from LockerHashLookupKey]
    F --> G{reverse_lookup row exists?}
    G -->|No| H[Record KV_MISS]
    H --> E
    G -->|Yes| I[ReverseLookup::get_partition_key]
    I --> J[HGet locker payload from Redis]
    J -->|Hit| K[Return Some Locker]
    J -->|Redis NotFound| L[Record KV_MISS]
    L --> E
    E --> M{Postgres row exists?}
    M -->|Yes| K
    M -->|No| N[Return None]
```

Tests:
- `cargo check --features "release"`
- `cargo check`
- `cargo fmt --check` (passes; stable rustfmt emits existing warnings for unstable rustfmt options)


src/storage/types.rs
SanchithHegde, line 102: Nit: Should we consider specifying a format for timestamp fields now that we're going to be serializing and deserializing this?

src/storage/types.rs
```diff
@@ -157,6 +176,15 @@ pub(crate) struct ReverseLookup {
     pub update_by: String,
```
SanchithHegde: Nit, unrelated: We could name this `updated_by` to keep it consistent.

src/storage/types.rs
SanchithHegde: Will this end up being serialized as a JSON array (something like `[123, 234, ..., 255]`)? Should we consider base64 encoding this when serializing (to reduce the size of the serialized value)?

src/storage/kv/partition_key.rs
```diff
@@ -1,15 +1,31 @@
...
+            Self::CombinationKey { combination } => f.write_str(&format!("{}", combination)),
             Self::Fingerprint { fingerprint_hash } => {
                 f.write_str(&format!("fingerprint_{}", hex::encode(fingerprint_hash)))
             }
+            Self::Locker {
+                merchant_id,
+                customer_id,
+                locker_id,
+            } => f.write_str(&format!("locker_{merchant_id}_{customer_id}_{locker_id}")),
         }
```
SanchithHegde, line 29: I feel we can slightly simplify this and avoid unnecessary allocations.


```suggestion
match self {
            Self::CombinationKey { combination } => f.write_str(combination),
Self::Fingerprint { fingerprint_hash } => {
write!(f, "fingerprint_{}", hex::encode(fingerprint_hash))
}
            Self::Locker {
                merchant_id,
                customer_id,
                locker_id,
            } => write!(f, "locker_{merchant_id}_{customer_id}_{locker_id}"),
}
```

src/storage/kv/resource.rs
```diff
@@ -58,14 +75,40 @@ pub(crate) trait KvResource:
...
+    + Sync
+    + Send
+    + Sized
+{
+    // type Error: error_stack::Context
+    //     + Send
+    //     + Sync
+    //     + 'static
+    //     + StorageErrorExt
+    //     + for<'a> From<&'a KvError>;
```
SanchithHegde: Nit: Can we remove commented code?

src/storage/kv/resource.rs
```diff
@@ -174,15 +273,93 @@ where
...
+) -> Result<M, ContainerError<M::Error>>
+where
+    M: KvReverseLookupResource,
+{
+    let scheme = decide(store, Op::Find).await;
+    let lookup_id = lookup_key.get_lookup_key();
+    match scheme {
+        StorageScheme::PostgresOnly => M::storage_find_by_lookup(store, &lookup_key).await,
+        StorageScheme::RedisKv => {
+            let key_str = match store.find_by_lookup_id(&lookup_id.lookup_id).await {
```
SanchithHegde, line 276: Do we fetch the reverse lookup value from the database here, and not Redis?

hrithikesh026: find_by_lookup_id internally would have its own kv implementation.

src/storage/types.rs
hrithikesh026: unfortunately, this column was already introduced as `update_by`. Should've been `updated_by`.

SanchithHegde: We haven't deployed this in any of our environments, neither have we released a stable version with this. We could rename it now I feel.

SanchithHegde (approved):

PR #183, opened 2026-07-14
feat(kv): add hash table kv support

hrithikesh026:
Summary:
- Adds KV support for the `hash_table` storage path behind the `kv` feature.
- Introduces the `hash_table` KV resource implementation, including entity typing, Redis partition-key derivation from `data_hash`, insert drainer query generation, and Postgres fallback reads.
- Routes hash insert/find operations through the generic KV helpers when KV is enabled while preserving the existing Postgres-only path for non-KV builds.
- Makes `HashTable` serializable/deserializable and extends `HashTableNew` with `created_at` so KV payloads and drainer inserts can round-trip correctly.
- Merged latest `main` and kept locker/fingerprint KV support compatible with the new hash-table partition key.

Tests:
- `cargo fmt`
- `cargo check --features kv`
- `cargo check`


src/storage/db.rs
```diff
@@ -226,34 +226,71 @@ impl super::HashInterface for Storage {
         &self,
         data_hash: &[u8],
```
SanchithHegde, line 227: We should consider having this as a `Secret<_>` instead.

SanchithHegde (approved):

PR #187, opened 2026-07-22
feat(kv): validate runtime storage transitions

hrithikesh026:
## Summary
- validate runtime KV mode transitions and ignore unsupported switches
- keep KV enablement global, refresh it from runtime config, and only move to enabled after Redis health validation succeeds
- fail server startup when Redis is configured but the Redis connection cannot be established
- track read-replica enablement globally and probe replica availability only while enabling replica reads
- expose runtime config status for KV and read-replica state visibility

### Transition from Disabled to SoftKill failed.
image
image

- `cargo test --features kv kv::scheme::tests`
- `cargo check --features kv`

### DIFF
--- a/src/runtime_config.rs
+++ b/src/runtime_config.rs
@@ -29,6 +29,22 @@ enum RuntimeConfigState {
     },
 }
 
+#[derive(Debug, serde::Serialize)]
+pub struct RuntimeConfigStatus {
+    pub status: RuntimeConfigStatusKind,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub config: Option<serde_json::Value>,
+}
+
+#[derive(Debug, serde::Serialize)]
+#[serde(rename_all = "snake_case")]
+pub enum RuntimeConfigStatusKind {
+    Disabled,
+    NotFetched,
+    Available,
+    Invalid,
+}
+
 /// Fetches the runtime-config endpoint on a schedule and serves the last-known-good body.
 pub struct RuntimeConfigManager {
     state: RuntimeConfigState,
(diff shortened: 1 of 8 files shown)


src/storage.rs
```diff
@@ -192,17 +216,47 @@ impl Storage {
         self.replica_pg_pool.is_some()
     }
 
-    /// Returns `true` when both a replica pool exists and runtime config allows replica reads.
+    pub async fn runtime_config_status(&self) -> StorageRuntimeConfigStatus {
+        StorageRuntimeConfigStatus {
+            runtime_config: self.runtime_config_manager.status().await,
+            storage: StorageRuntimeConfigState {
+                use_replica: self.use_replica.load(Ordering::Acquire),
```
SanchithHegde: Any specific reason why we're using `Acquire`, instead of something like `SeqCst`?

hrithikesh026: SeqCst is not really required here. Since it is a simple load() call.

src/runtime_config.rs
```diff
@@ -118,6 +134,38 @@ impl RuntimeConfigManager {
...
+            return RuntimeConfigStatus {
+                status: RuntimeConfigStatusKind::NotFetched,
+                config: None,
+            };
+        };
+
+        match serde_json::from_str(raw) {
+            Ok(config) => RuntimeConfigStatus {
+                status: RuntimeConfigStatusKind::Available,
+                config: Some(config),
```
SanchithHegde, line 157: Are we not going to check if the JSON value here matches our expected format / structure? Won't this possibly overwrite the runtime config with an invalid value that we may not be able to deserialize later?

src/storage.rs
```diff
@@ -220,11 +274,40 @@ impl Storage {
...
+        ) {
+            match &self.redis {
+                Some(redis) => redis.test().await.is_ok(),
+                None => false,
+            }
+        } else {
+            false
+        };
+
+        let mut current_state = self.kv_state.write().await;
```
SanchithHegde: Would we end up holding a write lock on every call to `kv_settings()`?

hrithikesh026: nice catch. I should change it. To lock only when there is a state change.

src/runtime_config.rs
hrithikesh026: This endpoint simply fetched the cached runtime config and returns it. it doesn't override the existing cached value.

SanchithHegde (approved):

PR #188, opened 2026-07-22
chore(kv): refactor wrapper delete behaviour

hrithikesh026:
## Summary
- Replace the generic enum-style KV wrapper with a backend-agnostic `KvBehaviour` interface.
- Introduce `KvBackend` as the storage-owned KV backend dispatcher and `RedisBackend` as the Redis implementation.
- Maintain a tenant-aware Redis backend inside `Storage` instead of constructing a new `RedisBackend` for each KV call.
- Add tenant-aware Redis key prefixing via `TenantAwareRedisStore`.
- Store delete tombstones in Redis so deleted records are treated as not found while still allowing re-insert over tombstones.
- Use Redis `WATCH`/`MULTI` for tombstone-aware conditional inserts.
- Emit `kv.transaction.retry.count` and a warning log when optimistic Redis transactions are retried.
- Route Redis health checks through `Storage::get_redis_store()`.
- Add env parsing support for comma-separated `redis.cluster_urls`.
- Require tenant `redis_key_prefix` when Redis is enabled and document the prefixed drainer stream format.

## Testing
- `cargo check --features kv`


Cargo.toml
```diff
@@ -60,6 +60,7 @@ config = "0.14.1"
 serde_path_to_error = "0.1.16"
 error-stack = "0.5.0"
 futures-util = "0.3.30"
+fred = { version = "8.0.6", optional = true, features = ["metrics", "partial-tracing", "subscriber-client"] }
```
SanchithHegde, line 63: Why are we directly using `fred` here? Do we have an option to use it via `redis_interface` instead?

src/storage/kv/wrapper.rs
```diff
@@ -1,33 +1,37 @@
...
-use crate::{logger, observability::metrics};
+use crate::{config::KvConfig, logger, observability::metrics, storage::redis as redis_store};
 
 /// Drainer-entry `request_id`: log-only, empty (not threaded), kept for wire-format parity.
 const REQUEST_ID: &str = "VAULT_CONSTANT_REQUEST_ID";
+const KV_TRANSACTION_MAX_RETRIES: usize = 3;
+
+#[derive(Debug, serde::Deserialize, serde::Serialize)]
+#[serde(untagged)]
+enum KvStoredValue<T> {
```
SanchithHegde, line 20: Nit: Can we preferably avoid untagged deserialization?

src/storage/kv/wrapper.rs
```diff
@@ -89,133 +93,392 @@ impl<T> BridgeRedis<T> for Result<T, error_stack_04::Report<RedisError>> {
...
+    Find,
+    Update,
+    Delete,
 }
 
-impl<T> std::fmt::Display for KvOperation<'_, T>
-where
-    T: serde::Serialize + Debug,
-{
+impl std::fmt::Display for KvOperationKind {
```
SanchithHegde: Nit: Can derive `strum::Display`, we already have that dependency.

src/storage/kv/wrapper.rs
```diff
@@ -89,133 +93,392 @@ impl<T> BridgeRedis<T> for Result<T, error_stack_04::Report<RedisError>> {
...
+        config: KvConfig,
+    ) -> Self {
+        Self::Redis(RedisBackend::new(redis, config))
+    }
+}
 
-    let ttl = store.ttl_for_kv();
+#[derive(Clone)]
+pub(crate) struct RedisBackend {
+    redis: Option<redis_store::TenantAwareRedisStore>,
```
SanchithHegde: If we are constructing a Redis-based backend itself, why would this field be optional?

src/storage/kv/wrapper.rs
```diff
@@ -89,133 +93,392 @@ impl<T> BridgeRedis<T> for Result<T, error_stack_04::Report<RedisError>> {
...
+    fn get_redis_conn(&self) -> error_stack::Result<Arc<RedisConnectionPool>, RedisError> {
+        self.redis
+            .as_ref()
+            .map(|redis| redis.get_redis_conn())
+            .ok_or_else(|| error_stack::Report::new(RedisError::RedisConnectionError))
+    }
 
-                redis_conn
-                    .set_hash_fields(&key.into(), vec![(value.0, serialized)], Some(ttl.into()))
+    async fn insert_if_absent_or_tombstone(
```
SanchithHegde, line 183: Can we have more logs in this method, for better visibility into the various scenarios that could happen?

src/storage/kv/wrapper.rs
```diff
@@ -89,133 +93,392 @@ impl<T> BridgeRedis<T> for Result<T, error_stack_04::Report<RedisError>> {
...
                     // drainer entry — accepted per eventual-consistency model; alert on
                     // KV_FAILED_TO_PUSH_TO_DRAINER.
-                    push_to_drainer_stream::<S>(store, query, partition_key).await?;
-                    Ok(KvResult::HSetNx(result))
-                } else {
-                    Ok(KvResult::HSetNx(HsetnxReply::KeyNotSet))
+                    push_to_drainer_stream::<V>(self, query, partition_key).await?;
+                    Ok(KvInsertResult::Inserted)
                 }
+                KvInsertResult::AlreadyExists => Ok(KvInsertResult::AlreadyExists),
```
SanchithHegde, line 413: Do we get metrics indicating whether the insert happened successfully or if it didn't because the key already existed?

Cargo.toml
hrithikesh026: a few traits from fred are being used for calling direct redis command function in insert_if_absent_or_tombstone function. Like watch, unwatch, multi, exec etc.

src/storage/kv/wrapper.rs
hrithikesh026: Can be done. But what is the concern with untagged deserialization???

SanchithHegde: From what I remember, untagged deserialization may require parsing the input recursively, until the enum variant is decided based on field names. If I remember correctly, this has the possibility of causing recursions (in the worst case).

src/storage/kv/wrapper.rs
```diff
@@ -272,17 +299,34 @@ impl RedisBackend {
...
+        );
+        record_kv_insert_result("retry_exhausted");
         Err(RedisError::SetHashFieldFailed.into())
     }
 
     fn is_tombstone(value: &[u8]) -> bool {
         matches!(
-            serde_json::from_slice::<KvTombstone>(value),
-            Ok(tombstone) if tombstone == Self::TOMBSTONE_VALUE
+            serde_json::from_slice::<KvStoredValue<de::IgnoredAny>>(value),
```
SanchithHegde, line 310: Is `IgnoredAny` being used here intentional?

src/storage/kv/wrapper.rs
```diff
@@ -272,17 +299,34 @@ impl RedisBackend {
...
+            Ok(KvStoredValue::Tombstone(tombstone)) if tombstone == Self::TOMBSTONE_VALUE
         )
     }
 }
 
+fn record_kv_insert_result(result: &'static str) {
```
SanchithHegde: Nit: Would it be possible to pass `resource` (`entity`, `merchant`, `locker`, `vault`, etc.) as an attribute to the metric? If not, it's okay.

Cargo.toml
SanchithHegde: Preferably, we should have them in `redis_interface` though.

src/storage/kv/wrapper.rs
hrithikesh026: Yeah. For the purpose of this function, I don't really need to spend any resource to trying to deserialize to serde::Value.

SanchithHegde (dismissed):

jarnura (dismissed):

SanchithHegde (approved):

PR #196, opened 2026-08-12
fix(kv): use secondary keys for hash fields

hrithikesh026:
## Summary
- add `SecondaryKey` support for KV resources and derive it from each resource primary key
- use `PartitionKey` as the Redis hash key and `SecondaryKey` as the Redis hash field key
- derive insert partition/secondary keys internally from `PrimaryKeyType`
- include both partition and secondary keys in KV duplicate/error context

## Testing
- `cargo check --features "release"`

### DIFF
--- a/src/storage/kv/resource.rs
+++ b/src/storage/kv/resource.rs
@@ -256,21 +256,34 @@ pub(crate) trait KvSecondaryLookupResource:
 }
 
 pub(crate) enum InsertConflictKey {
-    PartitionKey(String),
-    LookupKey(String),
+    PartitionKey {
+        partition_key: String,
+        secondary_key: String,
+    },
+    LookupKey {
+        partition_key: String,
+        secondary_key: String,
+    },
 }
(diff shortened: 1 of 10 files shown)


src/storage/kv/resource.rs
```diff
@@ -218,21 +256,34 @@ pub(crate) trait KvSecondaryLookupResource:
...
         match self {
-            Self::PartitionKey(key) | Self::LookupKey(key) => key,
+            Self::PartitionKey {
+                partition_key,
+                secondary_key,
+            }
+            | Self::LookupKey {
+                partition_key,
+                secondary_key,
+            } => kv_key_context(partition_key, secondary_key),
```
SanchithHegde, line 270: If this is being used only for logging, we can consider logging the structs / enums directly.

SanchithHegde (dismissed):

SanchithHegde (approved):

PR #199, opened 2026-08-19
feat(runtime-config): per-tenant Postgres source of truth with Redis cache

hrithikesh026:
## Summary

Replaces HTTP-polling runtime configuration with a per-tenant store backed by Postgres as the source of truth and a read-through Redis cache. Removes polling, adds an authenticated admin endpoint for updates, and tolerates Redis failures (bounded by TTL).

- **Source of truth**: new `configs` table per tenant schema (migration `2026-08-17-120000_create_configs_table`)
- **Cache**: per-tenant Redis entry with TTL; on miss or error, falls back to Postgres and re-populates best-effort
- **Bootstrap**: on startup, if the row is missing, seeds `{ "use_replica": false, "enable_kv": "disabled" }` and applies the resulting KV / replica transitions
- **Admin endpoint**: `POST /runtime-config` guarded by `x-internal-api-key`, plus existing `GET /health/runtime-config` for read-only checks
- **Config surface**: all `runtime_config.endpoint.*` TOML keys removed; the only knob is `runtime_config.admin_api_key` (env: `LOCKER__RUNTIME_CONFIG__ADMIN_API_KEY`)

## Admin endpoint

```
POST /runtime-config
Headers:
  x-tenant-id: <tenant>
  x-internal-api-key: <admin_api_key>
Body:
  {"value": {"use_replica": true, "enable_kv": "enabled"}}
```

Upserts Postgres (source of truth), invalidates the tenant's Redis cache key (`DEL` — TTL bounds staleness on failure), and applies KV / replica state transitions.

## Notable implementation points

- New `ConfigInterface` trait on `Storage` with `find_config` / `upsert_config`; `type Error = RuntimeConfigError` so callers get already-classified errors
- New `ConfigUpdate` (AsChangeset) restricts `ON CONFLICT` writes to `value` and `updated_at` — never rewrites `key` or `created_at`
- `impl_storage_error!` macro invocation for `RuntimeConfigError` maps raw diesel errors (`NotFound` / unique-violation / other) onto enum variants
- Runtime config manager holds no storage handle; `&Storage` is passed into each method so the same pool serves reads and writes

## Migration

Single squashed migration `2026-08-17-120000_create_configs_table`:
- `key VARCHAR(255) PRIMARY KEY`
- `value JSONB NOT NULL`
- `created_at / updated_at TIMESTAMP NOT NULL` (application-populated, no DB defaults)

## Test plan

- [ ] `cargo check --all-features` and `cargo clippy --all-features --lib` clean
- [ ] Boot with `runtime_config.mode = "enabled"` and no row: service seeds default and applies `enable_kv=disabled`, `use_replica=false`
- [ ] `POST /runtime-config` with valid key updates Postgres, invalidates Redis, transitions take effect immediately (no polling lag)
- [ ] `POST /runtime-config` with invalid key returns 401
- [ ] `GET /health/runtime-config` reflects the current state
- [ ] Redis down: reads fall back to Postgres; writes still succeed and are visible on next read
- [ ] Revert migration `down.sql` cleanly removes the table

## Breaking changes

All `runtime_config.endpoint.*` config keys are replaced with a single `runtime_config.admin_api_key`. Operators previously pointing at a remote config endpoint must migrate to the admin endpoint flow.

### DIFF
--- a/config/config.example.toml
+++ b/config/config.example.toml
@@ -153,21 +153,26 @@ background_metrics_collection_interval_secs = 15 # How often to collect metrics 
 # port = 9090                                      # Port the Prometheus HTTP server listens on
 # background_metrics_collection_interval_secs = 15 # How often to collect metrics in the background (in secs)
 
-# Runtime configuration endpoint
+# Runtime configuration stored locally in the per-tenant `configs` Postgres table
+# and read-through a per-tenant Redis cache. No polling.
+#
+# The runtime config source of truth is the per-tenant `configs` table
+# (created by migration 2026-08-17-120000_create_configs_table). The service
+# seeds a safe default (`{"use_replica":false,"enable_kv":"disabled"}`) at
+# startup when the row is absent.
+#
+# The only knob enabled mode adds is the admin API key that guards the
+# `POST /runtime-config` update endpoint:
+#   mode = "enabled"
+#   admin_api_key = "..."   # required when mode is "enabled"
+
 # [runtime_config]
 # mode = "enabled"
-# refresh_interval_seconds = 30
+# admin_api_key = "test_internal_admin_api_key"
 
-# Endpoint needs to be specified if mode is "enabled"
-# [runtime_config.endpoint]
-# base_url = "http://localhost:8090/configs"
-# api_key = "test_admin"
-# path = ""
-#
-# The endpoint at `{base_url}{path}` returns `{"key": "...", "value": "<config json string>"}`,
-# where `value` is a JSON string holding the flat config object, e.g.:
-#   {"key":"locker.configs","value":"{\"enable_kv\":\"enabled\",\"use_replica\":false}"}
-
-# Optional static headers sent on every runtime config pull request (omit entirely to send none)
-# [runtime_config.endpoint.headers]
-# accept = "application/json"
+# Update the runtime config via:
+#   curl -X POST http://<host>:<port>/runtime-config \
+#     -H "x-tenant-id: <tenant>" \
+#     -H "x-internal-api-key: <admin_api_key>" \
+#     -H "Content-Type: application/json" \
+#     -d '{"value":{"use_replica":true,"enable_kv":"enabled"}}'
--- a/migrations/2026-08-17-120000_create_configs_table/up.sql
+++ b/migrations/2026-08-17-120000_create_configs_table/up.sql
@@ -1,0 +1,11 @@
+-- Per-tenant runtime configuration store.
+-- Each tenant schema gets its own `configs` table; the application seeds
+-- a default `locker_runtime_config` row at startup when one is missing.
+-- Both `created_at` and `updated_at` are populated by the application on
+-- insert/update — no database defaults.
+CREATE TABLE IF NOT EXISTS configs (
+    key        VARCHAR(255) PRIMARY KEY,
+    value      JSONB        NOT NULL,
+    created_at TIMESTAMP    NOT NULL,
+    updated_at TIMESTAMP    NOT NULL
+);
--- a/src/config.rs
+++ b/src/config.rs
@@ -5,7 +5,9 @@ use std::{
 };
 
 use error_stack::ResultExt;
-use hyperswitch_masking::{ExposeInterface, PeekInterface, Secret};
+#[cfg(feature = "redis")]
+use hyperswitch_masking::PeekInterface;
+use hyperswitch_masking::{ExposeInterface, Secret};
 #[cfg(feature = "redis")]
 use hyperswitch_redis_interface::RedisSettings;
 
@@ -42,6 +44,7 @@ pub struct GlobalConfig {
     pub external_key_manager: ExternalKeyManagerConfig,
     #[cfg(feature = "redis")]
     pub redis: Option<RedisSettings>,
+    #[cfg(feature = "redis")]
     #[serde(default)]
     pub runtime_config: RuntimeConfig,
     #[cfg(feature = "kv")]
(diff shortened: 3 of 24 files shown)


config/config.example.toml
```diff
@@ -153,21 +153,26 @@ background_metrics_collection_interval_secs = 15 # How often to collect metrics
...
+# `POST /runtime-config` update endpoint:
+#   mode = "enabled"
+#   admin_api_key = "..."   # required when mode is "enabled"
 
-# Optional static headers sent on every runtime config pull request (omit entirely to send none)
-# [runtime_config.endpoint.headers]
-# accept = "application/json"
+# [runtime_config]
+# mode = "enabled"
+# admin_api_key = "test_internal_admin_api_key"
```
SanchithHegde: Can we keep the naming consistent please: either we use "admin API key" everywhere (config name, header name, in code) or we use "internal API key" everywhere, not a mix of both.

migrations/2026-08-17-120000_create_configs_table/up.sql
```diff
@@ -0,0 +1,11 @@
+-- Per-tenant runtime configuration store.
+-- Each tenant schema gets its own `configs` table; the application seeds
+-- a default `locker_runtime_config` row at startup when one is missing.
+-- Both `created_at` and `updated_at` are populated by the application on
+-- insert/update — no database defaults.
```
SanchithHegde: Can we please simplify the over-explanatory comments?

config/config.example.toml
```diff
@@ -153,21 +153,26 @@ background_metrics_collection_interval_secs = 15 # How often to collect metrics
...
-#   {"key":"locker.configs","value":"{\"enable_kv\":\"enabled\",\"use_replica\":false}"}
+# The runtime config source of truth is the per-tenant `configs` table
+# (created by migration 2026-08-17-120000_create_configs_table). The service
+# seeds a safe default (`{"use_replica":false,"enable_kv":"disabled"}`) at
+# startup when the row is absent.
+#
+# The only knob enabled mode adds is the admin API key that guards the
+# `POST /runtime-config` update endpoint:
+#   mode = "enabled"
+#   admin_api_key = "..."   # required when mode is "enabled"
```
SanchithHegde: Same here, can we simplify/clean up the over-explanatory comments? 

We seem to be documenting the API contract twice as well, should we even be documenting the API in the sample config documentation though?

src/error.rs
```diff
@@ -70,6 +70,35 @@ pub enum StorageError {
...
     #[error("Read replica pool is not configured")]
     ReplicaPoolNotConfigured,
+    #[error("Storage initialization failed: {0}")]
+    InitializationError(&'static str),
+}
+
+/// Errors from the runtime-config manager (`configs` table + Redis cache).
+#[cfg(feature = "redis")]
+#[derive(Debug, thiserror::Error)]
+pub enum RuntimeConfigError {
```
SanchithHegde: Why is this behind the `redis` feature flag though?

src/error.rs
```diff
@@ -70,6 +70,35 @@ pub enum StorageError {
...
+    /// The runtime-config payload could not be (de)serialized.
+    #[error("Invalid runtime config value: {0}")]
+    InvalidValue(String),
+    /// Illegal `enable_kv` state transition for the currently persisted state.
+    #[error("Invalid KV state transition: {0}")]
+    InvalidStateTransition(String),
```
SanchithHegde: This seems to be tightly coupling the runtime config framework with KV. But is that the expected behavior though?

src/error.rs
```diff
@@ -70,6 +70,35 @@ pub enum StorageError {
...
+    /// `use_replica: true` was requested but no read replica pool is configured.
+    #[error("Cannot enable use_replica: no read replica is configured")]
+    NoReplicaConfigured,
+    /// `use_replica: true` was requested but the read replica is unreachable.
+    #[error("Cannot enable use_replica: read replica is unreachable")]
+    ReplicaUnreachable,
```
SanchithHegde: Same here, this seems to tightly couple the runtime config framework with very specific use-cases, in this case usage of replica.

src/routes/health.rs
```diff
@@ -28,10 +28,14 @@ where
...
 pub fn serve() -> axum::Router<Arc<GlobalAppState>> {
-    axum::Router::new()
+    let router = axum::Router::new()
         .route("/", get(health))
-        .route("/diagnostics", get(diagnostics))
-        .route("/runtime-config", get(runtime_config_status))
+        .route("/diagnostics", get(diagnostics));
+
+    #[cfg(feature = "redis")]
+    let router = router.route("/runtime-config", get(runtime_config_status));
```
SanchithHegde: In my opinion, if we have a separate set of routes (`/runtime-config/*`) to handle runtime config, then the route to retrieve the value should also be under the same scope, not under `/health`.

src/routes/runtime_config.rs
```diff
@@ -0,0 +1,68 @@
...
+    pub value: serde_json::Value,
+}
+
+/// `POST /runtime-config`
+///
+/// Auth:
+///   - `x-tenant-id`        → tenant whose config table is updated
+///   - `x-internal-api-key` → shared secret — must match `runtime_config.admin_api_key`
+#[tracing::instrument(skip_all)]
+pub async fn update_runtime_config(
```
SanchithHegde: Don't we have the option to update/add the config for a specific key? Why are we required to provide the entire config again?

src/config.rs
```diff
@@ -378,7 +386,11 @@ impl GlobalConfig {
 
     pub fn validate(&self) -> error_stack::Result<(), error::ConfigurationError> {
         self.secrets_management.validate()?;
-        self.runtime_config.validate()?;
+        #[cfg(feature = "redis")]
+        {
+            self.runtime_config.validate()?;
+            self.validate_runtime_config_redis()?;
```
hrithikesh026: this can be merged into a single impl function.

PR #204, opened 2026-09-10
feat(fingerprint): add capability to return back plain response

kashif-m:
## Description

Adds an opt-in way for callers to receive the `/fingerprint` response as plain JSON instead of a JWE/JWS envelope.

A fingerprint response carries only a fingerprint id: no card data, no PII. Encrypting and signing it costs ~1.1ms on the vault, and the caller then spends about as much again decrypting and verifying it. The caller can now skip both.

**Behaviour**

- A request to a route ending in `/fingerprint` (`/cards/fingerprint`, `/api/v2/vault/fingerprint`) that carries `x-fp-response-encoding: plain` gets a plain JSON body. The response echoes the same header.
- Without the header, the response is encrypted by default.

Because the vault echoes the header, a caller decides how to decode from the response alone. That means the vault and its callers can be deployed in any order: a caller talking to an older vault simply receives an envelope.

**Changes**

- `src/middleware.rs`: `wants_plain_response` checks the route and the header; when both match, the middleware returns the plain body and echoes the header instead of encrypting.
- `src/storage/consts.rs`: adds the header name, its value, and the fingerprint path suffix.


su-shivanshmathur (dismissed):

jarnura (dismissed):

src/middleware.rs
```diff
@@ -36,14 +38,31 @@ where
...
 }
 
+/// Whether this request may receive a plain (unencrypted) response.
+///
+/// Only when the route is `/fingerprint` and the caller explicitly asked for it; without the
+/// header the response is encrypted as usual. A fingerprint response carries only a fingerprint
+/// id, so skipping response encryption for it is safe; the request payload is still decrypted and
+/// authenticated as usual.
+fn wants_plain_response(parts: &request::Parts) -> bool {
+    parts.uri.path().ends_with(consts::FINGERPRINT_PATH_SUFFIX)
```
SanchithHegde: This ends-with matching has the possibility of allowing plain response on any endpoint added on a later date, which also ends with `/fingerprint`. Should we consider explicitly matching against the complete paths instead?

kashif-m: We can extend it to other flows in future if needed; but for now we can hardcode it to the only required path 👍

src/middleware.rs
```diff
@@ -38,14 +38,9 @@ where
...
-/// Whether this request may receive a plain (unencrypted) response.
-///
-/// Only when the route is `/fingerprint` and the caller explicitly asked for it; without the
-/// header the response is encrypted as usual. A fingerprint response carries only a fingerprint
-/// id, so skipping response encryption for it is safe; the request payload is still decrypted and
-/// authenticated as usual.
+/// Whether the caller asked for a plain response on a route that may return one
 fn wants_plain_response(parts: &request::Parts) -> bool {
-    parts.uri.path().ends_with(consts::FINGERPRINT_PATH_SUFFIX)
+    parts.uri.path() == consts::V2_FINGERPRINT_PATH
```
SanchithHegde, line 43: This could have been an array of allowed paths/endpoints maybe, but sure, this works for now.

SanchithHegde (approved):

src/middleware.rs
kashif-m: In that case, we can make it generic too - based on a more generic header; but we have scoped it only to fingerprint operations to set the right expectations

We can cover those use cases in future

